PR #14, opened 2013-03-25
WIP: Add 'float' type option for ANTS computations

aghayoor:
The final goal of this project is to make antsRegistration program capable to
do its computations using 'float' precision instead of 'double'. It helps CPU
to do more processings per time unit when the float precision is enough
for our computations.

By this patch antsRegistration is made as a template over the type that can
accept 'float' or 'double'. The float precision is enabled when the
"--float" flag is used.
Although based on current version of ITK we cannot use such an option,
it can be used later when the registration framework of ITK allows us to use
float.

All ANTS tests are passed successfully: 
"100% tests passed, 0 tests failed out of 291".


aghayoor:
This patch has been updated and merged already.

PR #28, opened 2013-06-28
ENH: compareTwoCompositeTransforms is renamed and enhanced

aghayoor:
Previously compareTwoCompositeTransforms was written to compare only
two input composite tranforms.
I renamed this program to "compareTwoTransforms" and enhanced its
functionaliy, so it can compare several kind of transforms including
affine type tranforms and displacement field transform. Note that it
does not support BSpline type transforms.


aghayoor:
Abandoned! A new patch will be submitted in this regard.

PR #94, opened 2014-06-25
BUG: Remove const_cast wherever that was possible

aghayoor:
Based on some new changes to ITKv4, we can avoid using const_cast in
many places in ANTs. Therefore, I updated the ITK tag, and I removed
const_cast from everywhere that it could be avoided.


aghayoor:
This pull request is already merged.

PR #145, opened 2015-01-14
Add auto versioning

hjmjohnson:
@satra @ntustison @stnava 

In order to better support nipype, I have added an initial bit of logic for adding versioning information to ANTs.  I arbitrarily chose the MAJOR_VERSION to be 15.

NOTE: I only instrumented the antsRegistration tool.  Someone else will need to copy-n-paste the command line processing code for "--version" and "-v" command line options to the other programs.


armaneshaghi:
@hjmjohnson  Thanks, shall we put version to 2.1 to be included in the version coming soon?

stnava:
yes - agree w/that arman


brian

armaneshaghi:
I will merge this and add version to other programs as @hjmjohnson instructed.

hjmjohnson:
@armaneshaghi @stnava Changed patch to be version 2.1 as requested.  It should now be ready to merge.

satra:
@hjmjohnson - i think it would be good to leave a git trail following the minor version `2.1.0-git_short_hash`. this way for releases only the major/minor/patch needs to be updated, and you get the added benefit of always knowing the githash as well.

hjmjohnson:
@satra This is the way it is configured.  You must manually set the MAJOR,MINOR,PATCH versions, but the git_short_hash is auto populated by cmake during compilation.

satra:
@hjmjohnson - great

PR #149, opened 2015-01-15
ENH: compareTwoCompositeTransforms is enhanced for more tx types

aghayoor:
Previously compareTwoCompositeTransforms was written to compare only
two input composite tranforms.
I renamed this program to "compareTwoTransforms" and enhanced its
functionaliy, so it can compare several kind of transforms including
affine type tranforms and displacement field transform. Note that it
does not support BSpline type transforms.


aghayoor:
@hjmjohnson @stnava 
Please review this patch.

hjmjohnson:
@aghayoor Looks good to me.
@ntustison

aghayoor:
Hello all,

All tests are passed successfully after this patch. This is ready to be merged as soon as you confirm that.

Thanks,
Ali

PR #152, opened 2015-01-16
BUG: Fixed three test failures on dashboard

aghayoor:
Three antsRegistration tests were failed during the last few weeks due
to negligible difference between the result and baseline files. I
carefully inspected the differences and verified that is safe to fix the
failures by replacing the baseline files.


aghayoor:
Previous failures on dashboard:
http://testing.psychiatry.uiowa.edu/CDash/viewTest.php?onlyfailed&buildid=37978

Now, all ANTs tests are passed successfully:
100% tests passed, 0 tests failed out of 289

PR #153, opened 2015-01-16
FIX: ComposeMultiTransform always return 1 when argc > 3

lsqshr:
`ComposeMultiTransform` always returned 1 which stopped my nipype commandline (No error message was thrown and the result file looks good).

It seems that it always `return EXIT_FAILURE` regardless of the `if( is_parsing_ok )` condition. Could you guys pls double check this fix? Thanks!


ntustison:
Hey , thanks for doing this.

PR #171, opened 2015-03-09
BUG: ImageMath Canny edge detection gives 0 output

armaneshaghi:
I'm going to need this for a new feature in plot.antsImage in ANTsR for checking
registration where overlaying edge-detected image on another image gives a clear feeling of
how registration has worked :

 - Leaving thresholds to ITK defaults gives more stable
results.
 - We don't need rescaling because the output of the filter is
(on its own) between 0 and 1. 
 - Some cleaning up of the code, removing unnecessary variables, `image2`

### DIFF
--- a/Examples/ImageMath.cxx
+++ b/Examples/ImageMath.cxx
@@ -8505,42 +8505,32 @@ int CannyImage(      int argc, char *argv[])
     {
     sig = 0.5;
     }
-  PixelType lowerThreshold = 0.5;
+  PixelType lowerThreshold = -1.0;
   if( argc > argct )
     {
     lowerThreshold = atof(argv[argct]); argct++;
     }
-  PixelType upperThreshold = 1;
+  PixelType upperThreshold = -1.0;
   if( argc > argct )
     {
     upperThreshold = atof(argv[argct]); argct++;
     }
 
   typename ImageType::Pointer image = ITK_NULLPTR;
-  typename ImageType::Pointer image2 = ITK_NULLPTR;
   ReadImage<ImageType>(image, fn1.c_str() );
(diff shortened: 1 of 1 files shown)


stnava:
Thanks for fixing this -

stnava:
It would also be good if you could figure out how to set up reasonable
parameters for this detector, I haven't figured out how to get it to behave
in a natural way with Sigma etc.

armaneshaghi:
sure.

bkandel:
I don't quite follow the motivation for the addition to plot.ANTsR.  To
check registration, isn't it much simpler to just overlay two images?  I'm
not sure I see why using a Canny filter should be the default way to check
registration. I can see why one would want to do that, but it's not obvious
to me that this should be included in plot.antsImage.

2015-03-09 10:51 GMT-04:00 Arman Eshaghi <notifications@github.com>:

> sure.
>

armaneshaghi:
perhaps a matter of taste; I would rather to see statistical maps as
overlays, but check registration with edge detectors, similar to FSL.

Have to revert some commits on my own branch on ANTsR if you don't like
this.

stnava:
Shouldn't you make a wrapper function with application specific calls to
plot . ants image instead of directly modifying the function?

armaneshaghi:
checkReg.R ? does it work for you @bkandel?

bkandel:
That would be fine with me.  Depending on how widely used it becomes, it
may be worthwhile to put in "ANTsRMas" or whatever we call the "extended"
version of ANTsR.

2015-03-09 11:48 GMT-04:00 Arman Eshaghi <notifications@github.com>:

> checkReg.R ? does it work for you @bkandel?
>
>

PR #190, opened 2015-05-28
ENH: Addition of an install rule for target antsUtilities

fbudin69500:
When installing ANTs, target antsUtilities was not installed.


gdevenyi:
Thanks for the addition to the makefile, I didn't realize I was missing this tool!

PR #191, opened 2015-05-29
ENH: Make a member function to simplify use

hjmjohnson:
GetShrinkImageOutputInformation provides a consistent way to compute the
outputImage space for each level of a registration in a consistent way.
By always using the same reference image, we can ensure that the same
shrink results always are produced.


ntustison:
This is great.  Thanks Hans.

PR #202, opened 2015-07-06
Add support for writing output in minc-style .xfm  files

vfonov:
Added support for writing .xfm files to ANTS and  antsRegistration binaries. 
antsRegistration will have --minc option to make it produce .xfm files as output.

### DIFF
--- a/Examples/antsRegistration.cxx
+++ b/Examples/antsRegistration.cxx
@@ -473,6 +473,16 @@ static void antsRegistrationInitializeCommandLineOptions( itk::ants::CommandLine
   }
 
   {
+  std::string description = std::string( "Use MINC file formats for transformations." );
+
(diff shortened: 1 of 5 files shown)


stnava:
hi @vfonov  --- obviously, we'd like to merge this but we would also like to check it first ... but our travis builds are failing.  will get back to this once we get the core ants build working.

stnava:
got builds working so let's see how this goes on travis

vfonov:
it's because latest version of libminc has niftiio library inside of it , with ZLIB somehow disabled. So, when ITK is built with MINC library it recieves two versions of niftiio , one with support for .gz and another one without. I recently fixed it in libminc: https://github.com/BIC-MNI/libminc/commit/dcb93a5f50d2038fbe723874bb1407ef86363cdb , but it didn't make it into ITK yet.

PR #209, opened 2015-07-21
Fixed writing transformations per stage in minc-style .xfm files

vfonov:
also fixed output in case of unsupported transformation type, fixes stnava/ANTs#207


vfonov:
I am going to include the changes mentioned in #210 , so don't merge

PR #226, opened 2015-09-15
Fix initialization per stage

aghayoor:


aghayoor:
@hjmjohnson : Please review this pull request.
As I noticed, the combination of transforms is very prone to numerical precision. Whenever "antsRegistrationTest_initialize_transforms_per_stage" is run, generates slightly! different results.
Therefore, sometimes "antsRegistrationTest_initializePerStage_ComparisonTesting" test is fails, sometimes it is passed even if I choose very loose tolerance like 1e-1!

hjmjohnson:
@aghayoor Testing values of a transform is not going to be very reproducible due to the iterative multiprocessor floating point accumulation problem.

hjmjohnson:
The test should always pass.  Even if the test is not very precise.

aghayoor:
@hjmjohnson :
Pull request is updated and the added two tests always pass.

PR #236, opened 2015-11-02
cmake: fixes for GITDIR-NOTFOUND

ignatenkobrain:
1. We will remove -gGITDIR-NOTFOUND because it is awful to see
   this in project version.
2. Do not include file which not exists. There is no need to
   configure it because it will not say anything due to missing
   information from git.
3. Remove unused ProjectSourceVersion.cmake* from top dir as it
   is not used anywhere.

Reference: https://github.com/stnava/ANTs/issues/235
Reference: https://github.com/stnava/ANTs/issues/213
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>


stnava:
i'll merge - hope for support if this leads to other issues down the line.   thanks for your help.

ignatenkobrain:
Sure, I will help anytime! Just cc me in github issue =)

PR #253, opened 2015-11-23
WIP: Return exit codes

cookpa:
Some changes to make exit codes get passed to the shell. Also replacing things like `return(0)` with `return(EXIT_SUCCESS)`.


cookpa:
As before, I left programs alone if they were not trivial to fix. But most should return correct values on failure now.

ntustison:
So is this patch ready to merge?

cookpa:
It says checks pending but in terms of code, it's ready.

PR #277, opened 2016-02-06
Normalize denoised image to input range

hjmjohnson:
@ntustison Nick,  The denoising program is very nice, but the output images where on a vastly different range than the input images.  This was pretty easy to remedy by simply applying a linear intensity transfer function such that the intensity rescaled output is most like the input image.

This patch hardcodes to always do the rescaling (a computationally cheap operation).  It should be easy to add a command line option for turning this on/off, but I'm not fluent in the ants command line option behaviors.


ntustison:
Looks good.  As soon as the checks complete I'll go ahead and merge.

hjmjohnson:
@ntustison Thanks!

Hans

PR #287, opened 2016-03-07
antsMultivariateTemplateConstruction2 not prefixing SGE jobs correctly

spinicist:
Hello,
I think I found and fixed a small bug where the job prefix was not being written out for SGE jobs in the 2nd stage.
I also took the opportunity to change the pattern used to strip file extensions from %.*.* to %%.* - the latter will match against .nii as well as .nii.gz, if anyone wants to change the script to output uncompressed Nifti files.
Thanks,
TOby

### DIFF
--- a/Scripts/antsMultivariateTemplateConstruction2.sh
+++ b/Scripts/antsMultivariateTemplateConstruction2.sh
@@ -1313,14 +1313,12 @@ while [[ $i -lt ${ITERATIONLIMIT} ]];
                 indir=`pwd`
               fi
             IMGbase=`basename ${IMAGESETARRAY[$l]}`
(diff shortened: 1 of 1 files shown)


ntustison:
Does it run successfully on this [example](https://github.com/ntustison/TemplateBuildingExample/blob/master/BrainSlices/templateCommandMultivariateBSplineSyN.sh)?

spinicist:
Hello Nick,
I assume you mean what happens if I run that script with my changes? I just downloaded and ran it. It appears to complete successfully, producing a template0.nii.gz and template1.nii.gz file. Do you need any more checks?
Toby

ntustison:
Yep, that's all I meant.  Thanks for doing that.  I'll go ahead and merge the patch.  Thanks.

PR #289, opened 2016-03-11
ENH: Masking operations uniform use nonzero values

hjmjohnson:
When applying a masking operation, any non-zero value
in the image should be considered foreground.  This is
the behavior of the itkMaskImageFilter, and the
itkSpatialImageMaskObject filters.  This makes using
the same image (even an label map image with multiple
labels) appropriate across different uses (i.e. registration
and segmentation).

In all uses inside ANTs, this had been hard coded to the
forground being equal to a value of 1, and no cases were found
that allowed for changing the mask match value.  For masks that
have only values of zero and 1, backwards compatibility is preserved.

For uses where label maps or intensity images were used, this
changes behavior, but in a way that makes ants tools consistent with
other ITK based behaviors.

### DIFF
--- a/Examples/Atropos.cxx
+++ b/Examples/Atropos.cxx
@@ -416,7 +416,7 @@ int AtroposSegmentation( itk::ants::CommandLineParser *parser )
                                                            segmenter->GetPriorLabelImage()->GetLargestPossibleRegion() );
         for( ItM.GoToBegin(), ItP.GoToBegin(); !ItM.IsAtEnd(); ++ItM, ++ItP )
           {
-          if( ItM.Get() == segmenter->GetMaskLabel() && ItP.Get() == 0 )
+          if( ItM.Get() != itk::NumericTraits<LabelType>::ZeroValue() && ItP.Get() == 0 )
             {
             if( verbose )
               {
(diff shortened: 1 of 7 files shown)


hjmjohnson:
@ntustison   Not yet tested.  Running long tests on my computer now.

ntustison:
okay, great.  I'll be back in a bit after running some errands.

hjmjohnson:
98% tests passed, 5 tests failed out of 296

Total Test time (real) = 381.41 sec

The following tests FAILED:
	 16 - antsRegistrationTest_SimilarityRotationNoMasks (Failed)
	 17 - antsRegistrationTest_SimilarityScaleMasks (Failed)
	 18 - antsRegistrationTest_SimilarityScaleNoMasks (Failed)
	 19 - antsRegistrationTest_SimilarityTranslationRescaleNoMasks (Failed)
	 20 - antsRegistrationTest_SimilarityTranslationRescaleNoMasks_Float (Failed)
Errors while running CTest


Failing tests are not related to this patch set.  I don't think there are any tests that would have testing these changes though.

ntustison:
Thanks again Hans.

PR #290, opened 2016-03-14
ENH: Modernizing vcl_ to std::

hjmjohnson:
In all supported compilers, the need for vcl_ specialized
functions has been removed.  there is no longer a need
to have these overrides, and code is easier to read
and easier to maintain without these specializations.

The vcl_* definitions can now be greatly simplified.
After removing specializations for early non-conformant
c++ compilers the end result was that only the
std:: version of the functions could ever be
used by the compiler.

ITK_SCRIPT=ITK/Utilities/Maintenance/VCL_ModernizeNaming.py
SRC_BASE_DIR=$(pwd)
for ext in ".h" ".cxx" ".cpp" ".hxx" ".hpp" ".txx"; do
  find ${SRC_BASE_DIR} -type f -name "*${ext}" \         -exec python ${ITK_SCRIPT} {} \;
done

### DIFF
--- a/Examples/ANTS.cxx
+++ b/Examples/ANTS.cxx
@@ -232,7 +232,7 @@ private:
       }
 
     unsigned int numberOfLevels = static_cast<unsigned int>(
-        vcl_log( (double)maxSize / 32 ) / vcl_log( (double) 2 ) ) + 1;
+        std::log( (double)maxSize / 32 ) / std::log( (double) 2 ) ) + 1;
     std::string iterations( " -i " );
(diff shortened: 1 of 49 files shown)


ntustison:
Thanks for doing this, Hans.  I'll merge as soon as all the travis builds finish.

ntustison:
Hey Hans, do you delete these branches or do you prefer to keep them?

hjmjohnson:
I delete them once merged.

Hans

PR #293, opened 2016-03-17
Customization for specific tools installation

fbudin69500:
When integrating ANTs in a different project, one may only require a few tools that are part of ANTs' package. If ANTs was integrated using the CMake command Add_ExternalProject(), one could not install only a few specific tools using CPack and CPACK_INSTALL_CMAKE_PROJECTS.
With the new implementation, one can write the following line of code in their project:
`set(CPACK_INSTALL_CMAKE_PROJECTS "${CPACK_INSTALL_CMAKE_PROJECTS};${ANTS_DIR};ANTS;RUNTIME_antsRegistration;/")`
and this will only package antsRegistration in their project.
Additionally, this pull-request improve the settings of the project and allows the user to specify the directory in which the binaries and the scripts are installed (previously it was harcoded to `bin`) using the CMake variables 'BIN_INSTALL_DIR' and `SCRIPTS_INSTALL_DIR`.


ntustison:
This looks good but we should probably check with @stnava to make sure these cmake changes do not break ANTsR.  If he doesn't get to it, I can take a look this weekend.

fbudin69500:
Thanks.

PR #294, opened 2016-03-20
WIP: Implementation of Generic interpolator for multilabel images

gdevenyi:
Came across http://www.insight-journal.org/browse/publication/950 which is a generalization of the MultiLabel interpolator included in ITK.

It seems as though it's not in ITK proper but it might be useful for transforming multilabel images, so I took the code from the paper and implemented it in the options for ANTs.

My intention is to allow specification of the standard interpolator types as a an option to the GenericLabel interpolator to allow interpolation of label transforms. Right now hard-coded to linear as per paper suggesting it works the best under their limited test case.

Comments requested. Is this worth adding to ANTs? Or should this module make it into ITK proper?

I am not the original author of the GenericLabel interpolator code.


ntustison:
Thanks for this, Gabriel.  Before merging this, though, please move the generic label interpolator class from Examples/include to the ImageRegistration directory.

gdevenyi:
Moved the class to ImageRegistration and added a bit more to the help output.

Things still to add are the other interpolators as options to the interpolator type for GenericLabel, I would expect at least BSpline as well. Not sure it makes sense to duplicate Gaussian as it replicates MultiLabel, or Nearest since it's a no-op. Does sinc make any sense at all to expose to the GenericLabel methods?

hjmjohnson:
@gdevenyi THANKS!

PR #300, opened 2016-04-09
Fixed initialization error

hjmjohnson:
This patch set fixes a nasty bug where sporadic failures occurred in DenoiseImage (especially on Linux).

### DIFF
--- a/Utilities/itkAdaptiveNonLocalMeansDenoisingImageFilter.hxx
+++ b/Utilities/itkAdaptiveNonLocalMeansDenoisingImageFilter.hxx
@@ -18,6 +18,11 @@
 #ifndef itkAdaptiveNonLocalMeansDenoisingImageFilter_hxx
 #define itkAdaptiveNonLocalMeansDenoisingImageFilter_hxx
 
+//HACK:
+#include "antsUtilities.h"
+
+#include "ReadWriteData.h"
+
 #include "itkAdaptiveNonLocalMeansDenoisingImageFilter.h"
 
 #include "itkArray.h"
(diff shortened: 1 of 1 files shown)


hjmjohnson:
@ntustison @stnava 

Bug squashed!

ntustison:
Great.  Thanks Hans.

ntustison:
@hjmjohnson was (this section)[https://github.com/stnava/ANTs/blob/master/Utilities/itkAdaptiveNonLocalMeansDenoisingImageFilter.hxx#L344-L375] a problem in the code?  The logic seems different now.

ntustison:
Never mind.  I went to the original code and it looks like I forgot that conditional statement in the old code.

hjmjohnson:
@ntustison   I'm glad you saw the old code.  It was not a problem, I just ran arross the duplicated code when debugging.   I do think that the lines
https://github.com/stnava/ANTs/blob/master/Utilities/itkAdaptiveNonLocalMeansDenoisingImageFilter.hxx#L372
https://github.com/stnava/ANTs/blob/master/Utilities/itkAdaptiveNonLocalMeansDenoisingImageFilter.hxx#L373

Should probably be protected better for multi-threaded applications.   I did the code refactoring when I thought this might be the problem.  It was not, but I think the less redundant code is better.

Hans

ntustison:
@hjmjohnson 

Here's the difference in logic to which I was referring. Compare this 
https://github.com/stnava/ANTs/blob/5fa74b0e6cecb82c6cb80d7d92bf8e4548318b83/Utilities/itkAdaptiveNonLocalMeansDenoisingImageFilter.hxx#L353-L367
and this
https://github.com/stnava/ANTs/blob/5fa74b0e6cecb82c6cb80d7d92bf8e4548318b83/Utilities/itkAdaptiveNonLocalMeansDenoisingImageFilter.hxx#L390-L401

That's why I went back to Jose's original C implementation and found that the difference in logic was a mistake on my part.  

Regarding the multi-threading protections---I totally agree.  The joint fusion program has similar issues but the cost overhead of adding the couple things I tried (what is it called---fast mutex lock or something like that?  I'm clearly not an expert) didn't justify the seemingly unlikely multi-threading scenarios which would be problematic.

PR #312, opened 2016-04-25
BUG: cleanup function in bash scripts killed all matching processes

cookpa:
The "cleanup" function in the various bash scripts did something like

```
runningANTSpids=( `ps -C antsRegistration -C ImageMath| awk '{ printf "%s\n", $1 ; }'` )
```

This does a system-wide trawl of the matching commands (varies by script), including all users. It's unlikely that you could kill other people's jobs unless you're root, but you can kill jobs of your own that are unrelated to the script you're trying to stop. 

I replaced this line with

```
 runningANTSpids=$( ps --ppid $$ -o pid= )
```

which attempts to kill off child processes of the script. This includes the ps command itself, which will always finish before it can be killed, but it won't affect programs not launched from the script.

### DIFF
--- a/Scripts/antsIntroduction.sh
+++ b/Scripts/antsIntroduction.sh
@@ -276,28 +276,19 @@ dataCheck
 
 }
 
+
 cleanup()
-# example cleanup function
 {
-
-  cd ${currentdir}/
-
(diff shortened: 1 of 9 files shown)


ntustison:
Thx Phil.

PR #320, opened 2016-05-26
Switch ITK URL to github

gdevenyi:
ITK.org is 4-5x slower than github (likely due to github's cloud hosting)

Switching the URL substantially improves download times. It sill supports both git:// and https://


ntustison:
@stnava @hjmjohnson This seems like a good switch.  Do y'all have an opinion on this?

PR #323, opened 2016-05-27
Fix cmake config call for external VTK build

gdevenyi:
Variable type was missing, breaking config call.


gdevenyi:
Closing because bug fixed by https://github.com/stnava/ANTs/commit/2f09e62faabf65162e66bd4eed8f94599522f409

PR #330, opened 2016-06-10
Fixed floating point error in ImageMath:LabelStats

rbarnard:
Fixed an error in volume calculation in ImageMath's LabelStats function. Volume was being calculated by accumulation, which leads to error with floating point values. Replaced by direct multiplication of voxel volume by voxel count.

### DIFF
--- a/Examples/ImageMath.cxx
+++ b/Examples/ImageMath.cxx
@@ -10192,7 +10192,6 @@ int LabelStats(      int argc, char *argv[])
       PixelType label = It.Get();
       if(  label == currentlabel  )
         {
-        totalvolume += volumeelement;
         totalct += 1;
         if( valimage )
           {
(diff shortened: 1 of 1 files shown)


stnava:
thank you

rbarnard:
The previous implementation assumed a constant voxel size as well.

Currently, the voxel size is calculated by multiplying each component of the image's spacing (via the `GetSpacing` method of `itk::ImageBase`).  I don't see any methods in the `itk::Image` class (or its superclasses) that provide support for variable voxel sizes (per-voxel spacing). What's the appropriate way to get the size of a particular voxel from an `itk::Image`?

ntustison:
@rbarnard and @gdevenyi ,

I would leave the contribution as is.  Even if MINC supports irregular slice sampling, that's not something that's accounted for in many, if any, of the algorithms of ITK  and, therefore, wouldn't be propagated to ANTs.  Perhaps if that's something that is modified in core ITK (like image orientation a few years back), we can revisit.

rbarnard:
Agreed. Thanks for your input, @ntustison.

PR #335, opened 2016-06-28
Add link to benchmarks

gdevenyi:
I recently performed some fixed/moving benchmarks for walltime and memory of antsRegistration.

Thought they might be useful.


stnava:
thx!

PR #363, opened 2016-09-27
Allow for both VTK versions during build

gdevenyi:
Check the major version of VTK and find the OpenGL rendering component based on this.

Warning: this can still cause problems if VTK7 was built with the old rendering backend. I don't know how to do a "this component OR that component" test in CMake.


gdevenyi:
Closed in favour of #366

PR #372, opened 2016-10-25
Add MINC support for config with VTK on

gdevenyi:
With VTK on MINC was missing from config, even if enabled. This fixes this.


gdevenyi:
Looks like this passes :)

gdevenyi:
Passes Travis.

PR #379, opened 2016-11-01
Allow for larger kernel sizes in the case of high-resolution data

gdevenyi:
When blurring high resolution data with large blurs, sometimes the we run into the maximum kernel with of 32. Increase this to a larger number so that in those cases the kernel isn't truncated.

Has no impact to cases where blur kernel isn't truncated because this is just the maximum.

### DIFF
--- a/Examples/SmoothImage.cxx
+++ b/Examples/SmoothImage.cxx
@@ -38,6 +38,8 @@ int SmoothImage(int argc, char *argv[])
   typename dgf::Pointer filter = dgf::New();
   typename medf::Pointer filter2 = medf::New();
   bool usespacing = false;
+  float gaussianmaxerror = 0.01;
+  int gaussianmaxkernelwidth = 32;
   if( argc  >  5 )
     {
     usespacing = atoi(argv[5]);
@@ -46,6 +48,14 @@ int SmoothImage(int argc, char *argv[])
   if( argc  >  6 )
     {
     usemedian = atoi(argv[6]);
+    }
+  if( argc  >  7 )
+    {
+    gaussianmaxerror = atof(argv[7]);
+    }
+  if( argc  >  8 )
+    {
+    gaussianmaxkernelwidth = atoi(argv[8]);
     }
   if( !usespacing )
     {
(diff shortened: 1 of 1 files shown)


stnava:
i would prefer this to be a command line parameter.  this change, as is, will almost certainly have a negative impact on most smoothing operations ( in terms of compute time ).

gdevenyi:
Based on my reading of https://itk.org/Doxygen/html/classitk_1_1DiscreteGaussianImageFilter.html ``SetMaximumKernelWidth``, would only impact instances where the following warning occurs:
```
WARNING: In /opt/quarantine/ANTs/git/src/build/ITKv4-install/include/ITK-4.11/itkGaussianOperator.hxx, line 61
GaussianOperator (0x7fffdcae62b0): Kernel size has exceeded the specified maximum width of 32 and has been truncated to 33 elements.  You can raise the maximum width using the SetMaximumKernelWidth method.
```

This change would allow the the maximum kernel width to increase past 32 in order to achieve ``SetMaximumError(0.1)``. In cases where this error has no appeared, it will have no impact since a kernel width of < 32 already satisfied the error criterion.

Do you agree that's the case? Would you still like this to be an option?

ntustison:
The chosen parameters for the Gaussian kernel in ``SmoothImage`` and elsewhere have been there for quite awhile perhaps even predating ANTs proper.  I would rather not change those default values due to your particular usage regardless of how minor the consequences might seem to be.  Perhaps in the future we might change those default values but, for now, I would recommend simply permitting access to those values on the command line.  In addition to the kernel size, you could also permit access to the maximum error.

stnava:
yeah - good idea


brian

gdevenyi:
@ntustison great idea, I'll do that.

gdevenyi:
Also helps if I don't typo the code and break the build :-1:

gdevenyi:
Re-built patch as to expose options of max error and max gaussian kernel size with defaults of previous implementation.

gdevenyi:
Hurrah passes travis.

gdevenyi:
Any comments about this? This is now implemented the way you requested.

PR #384, opened 2016-11-07
Update ITK version to fix broken HFD5-deflate plugin

gdevenyi:
Fixes broken MINC with internal compression using deflate.

Changes in ITK versions:
https://github.com/InsightSoftwareConsortium/ITK/compare/2714cc1805f50504f5b9a60d0f62ffec8e73989...5aad8ac86968a12be74a31ef6493811c0ea13d05


gdevenyi:
Passes travis. This may fix other IO that uses compression inside (like hdf5 composite transforms?)

PR #453, opened 2017-05-28
ENH:  Modifying masking option.

ntustison:
Hey @stnava, I'd like to modify the available masking options in the SyN scripts.  Currently you set them up such that a single mask can be specified and it is applied to the last stage.  I'd like to open that up and allow for multiple masks to be specified.  Take a look and let me know if you think these changes are okay.

### DIFF
--- a/Scripts/antsRegistrationSyN.sh
+++ b/Scripts/antsRegistrationSyN.sh
@@ -66,21 +66,24 @@ Optional arguments:
      -i:  initial transform(s) --- order specified on the command line matters
 
      -t:  transform type (default = 's')
-        t: translation
-        r: rigid
-        a: rigid + affine
-        s: rigid + affine + deformable syn
-        sr: rigid + deformable syn
-        so: deformable syn only
-        b: rigid + affine + deformable b-spline syn
-        br: rigid + deformable b-spline syn
-        bo: deformable b-spline syn only
+        t: translation (1 stage)
+        r: rigid (1 stage)
+        a: rigid + affine (2 stages)
+        s: rigid + affine + deformable syn (3 stages)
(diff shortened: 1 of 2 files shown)


stnava:
that's fine - the issue i've seen in the past is this:

if a target mask is "small" e.g. a focal region of interest,

then the rigid / affine may not perform well.

so it's just a matter of alerting users to how the mask should be used.

potentially add comments to the help?

also - can you provide ( in help ) an example of how to pass several masks?

i assume for -sr  it would be -x mask1.nii.gz -x mask2.nii.gz  which would
correspond to

mask1 being used in syn and mask2 in rigid?


brian

ntustison:
Thanks Brian.   I'll add more to the help but this is what I have so far

```
-x:  mask(s) for the fixed image space.  Should specify either a single image to be used for
          all stages or one should specify a mask image for each "stage" (cf -t option).  If
          no mask is to be used for a particular stage, the keyword 'NULL' should be used
          in place of a file name.
```

plus I added the number of stages for each transform type in the help, i.e.,

```
     -t:  transform type (default = 's')
        t: translation (1 stage)
        r: rigid (1 stage)
        a: rigid + affine (2 stages)
        s: rigid + affine + deformable syn (3 stages)
        sr: rigid + deformable syn (2 stages)
        so: deformable syn only (1 stage)
        b: rigid + affine + deformable b-spline syn (3 stages)
        br: rigid + deformable b-spline syn (2 stages)
        bo: deformable b-spline syn only (1 stage)
```

So you're right about usage but one would have to do ``-x NULL -x maskForSyNStage.nii.gz`` to get the previous usage for the transform option ``-t sr``.  However, this mirrors the current ``antsRegistration`` call.

PR #455, opened 2017-06-01
ENH:  Refactoring MeasureImageSimilarity to use ITKv4 metrics.

ntustison:
Complete refactorization of MeasureImageSimilarity to use ants command line interface and ITKv4 metrics.

### DIFF
--- a/Examples/MeasureImageSimilarity.cxx
+++ b/Examples/MeasureImageSimilarity.cxx
@@ -1,398 +1,612 @@
-/*=========================================================================
-
-  Program:   Advanced Normalization Tools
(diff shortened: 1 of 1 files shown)


stnava:
any idea why travis fails on this?

ntustison:
Yeah, it looks like to fails on all those old ANTS_***_METRIC_* tests which call ``MeasureImageSimilarity``.  I completely changed the interface so those are obviously failing.  I was going to tackle that today.

stnava:
right - ok - will merge and then we can deal with next steps

PR #488, opened 2017-08-21
Set a default output prefix in antsMultivariateTemplateConstruction2.sh

jpeloquin:
The output prefix (-o) in `antsMultivariateTemplateConstruction2.sh` is currently compulsory, but there is a default set in the script.  To make this default work it is only necessary to set a corresponding default for the template name  This pull request makes this change.

This change, although not strictly necessary, is helpful because in the current implementation, if the user forgets to set the output prefix, the script seems to work at first.  It only fails after the first template iteration, which may take a very long time.


ntustison:
Thanks.

PR #497, opened 2017-09-06
BUG: fixed Cortical Thickness script when output suffix is not .nii.gz

aghayoor:


stnava:
@aghayoor - thanks much

aghayoor:
You're welcome! Thank you!

PR #501, opened 2017-10-03
Fix VTK/ITK dependencies order

gdevenyi:
Finally managed to figure out how to make cmake order the build properly.

Addresses https://github.com/ANTsX/ANTs/issues/466

Did this by

1) re-ordering the VTK and ITK dependencies in superbuild
2) Fixing a typo that caused VTK turning on to re-write the dependencies completely
3) Add VTK dependency to External ITK if turn on in SuperBuild


gdevenyi:
this passes Travis, would like an additional commit to enable VTK on Travis so all tools are tested?

ntustison:
Do you need anything other than merging this commit?

gdevenyi:
All this needs right now is a merge, I'm just wondering if the VTK bits should be turned on for Travis builds, so that that code gets tested regularly.

ntustison:
Does it time out with the VTK build on?  If not, then I don't see any reason why we shouldn't.

gdevenyi:
Will push a separate PR, see if the PR build times out :)

ntustison:
Great.  Thanks for doing that.

PR #521, opened 2017-11-22
COMP: Fix inconsistency of SetSigma types and overrides

hjmjohnson:
Remove unnecessary redundancy for the SetSigma() between the base class
and the derived class.

ANTs/Examples/SurfaceBasedSmoothing.cxx:7:
ANTs/Utilities/itkSurfaceImageCurvature.h:155:3: warning: 'SetSigma' overrides a member function but is not marked 'override'
  itkSetMacro(Sigma, float);
  ^
ANTs/Examples/SurfaceBasedSmoothing.cxx:71:3: note: in instantiation of template class
      'itk::SurfaceImageCurvature<itk::Image<float, 3> >' requested here
  ParamType::Pointer Parameterizer = ParamType::New();
  ^
ANTs/Utilities/itkSurfaceCurvatureBase.h:164:3: note: overridden virtual function is here
  itkSetMacro(Sigma, RealType);
  ^

### DIFF
--- a/Utilities/itkSurfaceImageCurvature.h
+++ b/Utilities/itkSurfaceImageCurvature.h
@@ -152,7 +152,6 @@ public:
   itkGetMacro(UseLabel, bool);
 
   itkSetMacro(kSign, float);
-  itkSetMacro(Sigma, float);
 
   itkSetMacro(Threshold, float);
 
@@ -260,7 +259,6 @@ private:
   NeighborhoodIteratorType m_ti2;
   bool                     m_UseLabel;
   float                    m_kSign;
-  float                    m_Sigma;
   float                    m_Threshold;
   float                    m_Area;
   RealType                 m_MinSpacing;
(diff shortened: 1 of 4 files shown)


Utilities/itkSurfaceImageCurvature.h
```diff
@@ -260,7 +259,6 @@ class SurfaceImageCurvature :
   NeighborhoodIteratorType m_ti2;
   bool                     m_UseLabel;
   float                    m_kSign;
-  float                    m_Sigma;
```
hjmjohnson: m_Sigma is already in the base class.  This shadow variable with the same name was confusing.  A minor side effect is that m_Sigma is of type "RealType" as defined in the base class.

Utilities/itkSurfaceImageCurvature.hxx
```diff
@@ -86,7 +86,6 @@ SurfaceImageCurvature<TSurface>
   m_UseLabel = false;
   m_kSign = -1.0;
   m_FunctionImage = ITK_NULLPTR;
-  m_Sigma = 1.0;
```
hjmjohnson: The base class defines m_Sigma, and also initializes it, so initialization is not needed in the derived class.

PR #527, opened 2017-12-10
Fix itk future deprecation errors

hjmjohnson:
These changes were identified as failures when deprecated code from ITK is removed.

### DIFF
--- a/Examples/CreateTiledMosaic.cxx
+++ b/Examples/CreateTiledMosaic.cxx
@@ -950,7 +950,7 @@ int CreateMosaic( itk::ants::CommandLineParser *parser )
           duplicator->SetInputImage( outputRgbSlice2 );
           duplicator->Update();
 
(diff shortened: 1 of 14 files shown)


hjmjohnson:
@ntustison  NOTE:  There is a pull request on ITK to make futureproofing easier.  Currently, ITK is not respecting the "-DITK_FUTURE_LEGACY_REMOVE:BOOL=ON" command line argument.  I did some contortions to make this work temporarily.  Hopefully, ITK will accept the configuration change into the master branch soon so that it will be easier to use this option.

ANTs also had a bug in it's use of ITK_FUTURE_LEGACY_REMOVE, so that has also been fixed.

hjmjohnson:
@ntustison Test pass except for 1 that timed out.

ntustison:
Okay, will merge this when I get home.  Apparently I can’t do it from my phone.

PR #545, opened 2018-03-05
Write Interval Volumes with Multiple Metrics

rbarnard:
Using the `--write-interval-volumes 1` argument of `antsRegistration` caused a segfault on stages that used multiple metrics. Updated the `WriteIntervalVolumes` method of the  `antsRegistrationOptimizerCommandIterationUpdate` class to work with both single-metric and multi-metric stages.


ntustison:
Thanks @rbarnard

PR #555, opened 2018-03-26
Fix for --write-interval-volumes filename padding

chrisadamsonmcri:
Having more than 100 iterations in one level causes naming for interval volumes to be sorted incorrectly. This fixes this.


ntustison:
Thanks

PR #557, opened 2018-03-28
COMP: Future proof code for null pointer robustness

hjmjohnson:
Prior to C++11 the NULL macro expanded to different
integer types on different compilers (sometimes size_t 0, usually
unsigned long 0).  C++11 provides a keyword nullptr to more robustly
identify a null pointer.  This robust null pointer (i.e. NOT an integer)
allows for more comprehensive type checking that is utilized in ITKv5.

As a consequence, implicit type conversion from integers (a.k.a NULL)
for smart pointers can cause compiler ambiguities in ITKv5 with C++11
enhanced pointer checking.

Using ITK_NULLPTR allows both ITKv4 and ITKv5 to compile and behave
as expected.

When building:
ITKv4 & CXX_STANDARD=98, ITK_NULLPTR -> NULL    (smartpointers can be initialized with integers like NULL)
ITKv4 & CXX_STANDARD=11, ITK_NULLPTR -> nullptr (prefer initializing smartpointers with nullptr, but NULL also works)
ITKv5 & CXX_STANDARD=11, ITK_NULLPTR -> nullptr (nullptr required for initializing smartpointers)

### DIFF
--- a/Examples/ANTS.cxx
+++ b/Examples/ANTS.cxx
@@ -107,7 +107,7 @@ int ANTSex(int argc, char *argv[])
 
 // entry point for the library; parameter 'args' is equivalent to 'argv' in (argc,argv) of commandline parameters to
 // 'main()'
-int ANTS( std::vector<std::string> args, std::ostream* /*out_stream = NULL*/ )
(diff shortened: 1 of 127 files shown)


ntustison:
Thanks Hans.  I had no idea that this wasn't consistent throughout.

PR #568, opened 2018-04-12
Improve transform filetype extension matching

### LINKED ISSUE #567: itkantsReadWriteTransform fails if filepath contains .mat, .txt, etc. anywhere
We ran into a confusing behavior where antsApplyTransforms would sometimes fail to read transform files, despite the file clearly existing with sufficient permissions. Upon some testing, we discovered that these same files could be read if we provided the relative, but not absolute, filepaths. The pattern turned out to be that the filepaths in question contained '.mat' inside the directory name. Specifically, we had an SGE queue named matlab.q, and /tmp files created for it were in a pattern of /tmp/*.matlab.q.*/. 

 I found that itkantsReadWriteTransform.h checks for any of these strings anywhere in the filepath: .h5, .hdf5, .hdf4, .mat, .txt, .xfm. If any of these is detected anywhere in the (absolute, if given) path name of a .nii.gz transform file, it will try to read it using itk affine transform readers. These then fail, and an error is thrown, rather than attempting to use the .nii transform reading functions.

The attached patch changes this logic so that the string matching is performed only on the extension rather than the entire filepath. This could certainly be implemented in many ways, but mine copied the logic from itkTxtTransformIO.hxx using itksys::SystemTools::GetFilenameLastExtension. itksys::SystemTools is already used elsewhere in itkantsReadWriteTransform.h, so this does not introduce an extra dependency.
[antsFilepathExtensionFix.txt](https://github.com/ANTsX/ANTs/files/1904770/antsFilepathExtensionFix.txt)

CGSchwarzMayo:
Fixes #567 upstream: itkantsReadWriteTransform fails if filepath contains .mat, .txt, etc. anywhere

### DIFF
--- a/Utilities/itkantsReadWriteTransform.h
+++ b/Utilities/itkantsReadWriteTransform.h
@@ -40,13 +40,15 @@ ReadTransform(const std::string & filename,
   // There are known tranform type extentions that should not be considered as imaging files
   // That would be used as deformatino feilds
   // If file is an hdf5 file, assume it is a tranform instead of an image.
-  if(    filename.find(".h5")   == std::string::npos
-      && filename.find(".hdf5") == std::string::npos
-      && filename.find(".hdf4") == std::string::npos
-      && filename.find(".mat")  == std::string::npos
-      && filename.find(".txt")  == std::string::npos
-      && filename.find(".xfm")  == std::string::npos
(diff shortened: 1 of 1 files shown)


ntustison:
Thanks for this.

PR #571, opened 2018-04-18
ENH: Allow compressing transform files

hjmjohnson:
When writing transforms to hdf5 format it is possible to use lossless
compression in ITKv5 to provide about 10% disk space savings.  For large
studies with many registrations, this can lead to significant disk space
savings.


gdevenyi:
Thanks @hjmjohnson this should also enable transparent compression inside MINC deformation field files.

Nice followup to https://github.com/ANTsX/ANTs/pull/305

PR #591, opened 2018-05-24
BUG: fixed a bug in CorticalThickness script

aghayoor:
Bug happens when the script checks if the output exists.


cookpa:
Thanks

PR #595, opened 2018-06-01
BUG: fix a bug in ImageSegmentation/antsAtroposSegmentationImageFilter.h

m-pilia:
Hello! The return type of that getter method was wrong, breaking compilation on gcc 8.1.0.


ntustison:
Thanks.

PR #599, opened 2018-06-11
Fix registration random seed

cookpa:
antsRegistration, antsAI, antsMotionCorr now take the "--random-seed" option. This lets the user set a non-zero seed, which is passed on to the registration method.

The order of precedence is:

1. Command line --random-seed
2. A non-null environment variable ANTS_RANDOM_SEED
3. Whatever the old default was (fixed for antsAI, system time for others)

Passing "--random-seed 0" will result in system time being used.


ntustison:
This is great @cookpa !  Much appreciated.

stnava:
Perfect

yarikoptic:
Cool. I wonder - what is the use case for having system time for a seed? (Ie how different from not seeding at all? Not that your would then try to match it up somehow)

cookpa:
The RNG requires a seed. If you don't specify one, then ITK will use the system time. 

https://itk.org/Doxygen/html/classitk_1_1Statistics_1_1MersenneTwisterRandomVariateGenerator.html

The idea is that you can have a different seed for each execution of the code. This can go wrong in parallel processing, because multiple jobs might start at the same time. But the impact in the registration context should be fairly minimal. For processes that rely heavily on good-quality randomness, it's more of a problem.

PR #628, opened 2018-08-28
Fix ANTSPATH needing trailing slash

gdevenyi:
Ran into these scripts throwing errors despite having ANTSPATH defined. Looks like some trailing slashes got dropped in the check for the functions.

This way, the check works as long as the path contains the needed commands.


ntustison:
@gdevenyi Thanks for this.

gdevenyi:
It was never required, or if it was the checks were all broken. I just made the checks uniform in the direction of the majority

PR #632, opened 2018-08-31
Normalize by the mean of the field for ITKN4

gdevenyi:
Add a feature to normalize the bias field by its mean so that overall the intensity isn't shifted up or down.

The existing linearly rescale intensities by the values in the mask didn't sit well with me, because specifying a mask also limits where the correction is applied.


gdevenyi:
I also fixed the help output for rescale-intensities to match its default value.

ntustison:
Hi @gdevenyi 

By default I'm averse to adding options so as to avoid feature creep and I don't think this feature is of sufficient necessity or convenience to include.  Normalizing by the mean of the bias field is somewhat of an arbitrary choice and is something that can also be done in a script with a couple lines.

gdevenyi:
Okay.

PR #639, opened 2018-09-16
Make .travis.yml more readable with YAML syntax "|".

junaruga:
Hello.
This PR is for refactoring of `.travis.yml`.
We can make `.travis.yml` more readable with YAML multi line syntax "|".

See https://stackoverflow.com/questions/38745696/how-to-use-multiline-command-in-script-with-yaml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,18 +22,32 @@ addons:
       - cmake-data
 
 before_install:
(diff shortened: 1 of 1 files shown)


ntustison:
Thanks for this.

Issue #634, opened 2018-09-04
Scripts: Greedy extension stripping removes parts of filenames


gdevenyi:
Greedy extension stripping such as https://github.com/ANTsX/ANTs/blob/master/Scripts/antsMultivariateTemplateConstruction2.sh#L1330

Will remove parts of files that have "." in them. Such as

``Subject01.extracted.nii.gz``
will be stripped to
``Subject01``
instead of
``Subject01.extracted``

sed/awk with a list of possible extensions to strip or splitting using IFS or other methods would avoid this.

coretemp:
You can better open an issue to replace the whole script. It should not be used by people.

gdevenyi:
Closed by #658

PR #658, opened 2018-11-02
Use pattern matching to remove extensions instead of greedy parameter expansion

gdevenyi:
Use pattern matching to avoid the bug in #634 

This will only strip nifti extensions, this may not be whats wanted, if so, happy to add more extensions to the list, however all the documentation, as well as some other hard-coded bits make this script currently not work properly on anything other than nifti.


ntustison:
Thanks @gdevenyi .

PR #662, opened 2018-11-07
ENH: Update for ITKv5 MultiThreader model

hjmjohnson:
Update to utilize the new MultiThreaderBase needed for
setting the number of threads.

### DIFF
--- a/Examples/itkTestMain.h
+++ b/Examples/itkTestMain.h
@@ -27,7 +27,7 @@
 #include <iostream>
 #include <fstream>
 #include "itkNumericTraits.h"
-#include "itkMultiThreader.h"
+#include "itkMultiThreaderBase.h"
 #include "itkImage.h"
 #include "itkImageFileReader.h"
 #include "itkImageFileWriter.h"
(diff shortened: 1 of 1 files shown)


ntustison:
Thanks @hjmjohnson

ntustison:
@hjmjohnson Great.  Thanks.  One question I was going to ask you concerned the new ITK "style"-(I don't know if that's the correct word) change based on, I'm guessing, C++11 standards.  I wrote some new classes ([e.g., simulated displacement fields](https://github.com/ANTsX/ANTs/blob/master/Utilities/itkSimulatedBSplineDisplacementFieldSource.h#L46)) and used this style but I'm wondering about changing existing classes.  Is there a script to make these changes or do we need to go in by hand and do it?

hjmjohnson:
Nick,

 

https://github.com/InsightSoftwareConsortium/ITK/tree/master/Utilities/ITKv5Preparation

 

I’ve got the infrastructure set up to do this already!  If you don’t mind, I’ll take a first pass. 

 

Hans

ntustison:
Awesome.  Thanks for the info and for doing the conversion.  This is great.

PR #673, opened 2018-11-25
ENH: Provide one version to rule them all

hjmjohnson:
CMake project command now supports setting the verison information
https://cmake.org/cmake/help/v3.10/command/project.html

This one place should provide a single location for configuring the
version of ANTs for packaging, library naming, and so versioning.


ntustison:
Thanks @hjmjohnson

PR #686, opened 2018-12-21
COMP: remove deprecated dynamic exception specifiers

ihnorton:
See https://discourse.itk.org/t/warning-dynamic-exception-specifications-are-deprecated-in-c-11/470

Allows to compile against ITK5 w/ `-std=c++17`.


ntustison:
Thanks for this.

Issue #688, opened 2018-12-22
cmdline Usage: - does it depict defaults?


yarikoptic:
Discovered while trying out LaplacianThickness bindings in nipype.  Came to realization that it seems that some people (including yours truly) possibly do not correctly treat Usage descriptions such as
```
Usage:   LaplacianThickness WM.nii GM.nii   Out.nii  {smoothparam=3} {priorthickval=5}  {dT=0.01}  use-sulcus-prior optional-laplacian-tolerance=0.001
```
From the first sight it kinda suggests that 
- I could specify in the command line those options as `option=value`, e.g. `LaplacianThickness WM.nii GM.nii   Out.nii  smoothparam=2 priorthickval=10 ` and thus may be even reorder them..  But that is not the case -- those are just positional arguments, so should be `LaplacianThickness WM.nii GM.nii   Out.nii  2 10`
- not clear why `optional-laplacian-tolerance=0.001` is just not `{laplacian-tolerance=0.001}` ?
- or why sulcus-prior is not `{sulcus-prior}`
- ALSO suggests that `=3` in `{smoothparam=3}` is to depict the default value (3), which is also not the case (code says `{smoothparam=3}`)

So -- what is the form `{param=value}` really is? Also other tools, including some other cmdline ones in ANTs, seems to use `[]` (not `{}`) do describe optional arguments.  But some times there is even `<>` seems to be used in ANTs:
```shell
$> ./RebaseTensorImage --help
Usage: RebaseTensorImage Dimension infile.nii outfile.nii <PHYSICAL/LOCAL/reference.nii.gz> 
```
how is that one different?  Could someone please clarify?

cookpa:
Unfortunately I don't think there's a unified style in ANTs. Square brackets as options, but that gets confusing with our use of brackets to denote groups of parameters. I think this is why, in many programs, braces are used for options.

In the example of `RebaseTensorImage`, you must specify one of the options in the angle bracket. I suppose this should use the pipe rather than slash to make that clearer. I think in most cases if you see angle brackets, it means a required argument.

yarikoptic:
my 1c: I think it could be an easy and super useful summer of code or "junior contributor" effort to unify and make it correspond to actual defaults.

PR #690, opened 2018-12-23
ENH: LaplacianThickness - make listed values to correspond to defaults and be uniform

yarikoptic:
Relates to my question #688 so may be I am doing something completely inappropriate ;)
Although I have modified that helper message, I am not 100% sure how to digest it since it doesn't really correspond to the code:
```c++
      float gmprob = gm->GetPixel(Iterator.GetIndex() );
      if( gmprob == 0 )
        {
        gmprob = 0.05;
        }
      float sprob = sulci->GetPixel(Iterator.GetIndex() );
      sprob = 1 / (1. + exp(-0.1 * (sprob - 0.5) / dosulc) );
      sulci->SetPixel(Iterator.GetIndex(), sprob );
```
and the message says
```
... 1/(1.+exp(-0.1*(laplacian-img-value-sulcprob)/0.01)) 
```

### DIFF
--- a/Examples/LaplacianThickness.cxx
+++ b/Examples/LaplacianThickness.cxx
@@ -984,11 +984,11 @@ private:
     {
     std::cout << "Usage:   " << argv[0]
              <<
-      " WM.nii GM.nii   Out.nii  {smoothparam=3} {priorthickval=5}  {dT=0.01}  use-sulcus-prior optional-laplacian-tolerance=0.001"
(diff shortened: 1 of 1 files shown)


ntustison:
Thanks @yarikoptic for this.  Sorry, some code is more consistent with the rest of the toolkit than others.

yarikoptic:
Glad to help
But could you also follow up on that comment about equation and to #688 just to get clear idea of what to aim for ;-)

PR #713, opened 2019-02-19
Fix some bracket globbing cases

bthyreau:
Tentative fix related to #712
File globbing can badly interact with ANTs bracket syntax in some cases. Here i suggest:
Some scripts do not need file-globbing; this is deactivated with "set -f".
Other scripts relies on file-globbing, so instead, potential error-cases are detected with preventive abort.

This PR only affects some scripts, as i didn't exhaustively check all of them.

### DIFF
--- a/Scripts/ants.sh
+++ b/Scripts/ants.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -f
 
 NUMPARAMS=$#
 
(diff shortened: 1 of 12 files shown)


gdevenyi:
My opinion is that globbing within pipelines/scripts is a dangerous anti-pattern since it can include similarly named files. This is particularly a problem for the ANTs scripts because they don't attempt to create a randomly named working directory or prefix their files with a random string (or the name prefix) to make things unique for a run, either of which would make globbing okay.

cookpa:
Closing this, will work on using white space to avoid glob traps

PR #723, opened 2019-02-26
Fix path to cmake config for vtk-8.2

gdevenyi:
Upgrade to vtk-8.2 didn't update the path to the cmake superbuild path.


gdevenyi:
Hi, any concerns with this? Right now the superbuild with VTK is broken.

cookpa:
Thanks, was failing for me also.

PR #754, opened 2019-04-09
PERF: Allow overwriting ABI/Optimization & Warning flags

hjmjohnson:
Provide a mechanism for changing the default warning
and ABI/OPTIMIZATION flags.

The default ABI/Optimization flag settings are
set to:
    -mtune=native # Tune the code for the computer used compile
                  # ITK, but allow running on generic cpu
                  # archetectures
    -march=corei7-avx # Use ABI settings to support corei7-avx
                      # (circa 2013 ABI feature sets)

The defaults can easily be overriden from the cmake comand line with:
  -DANTs_C_OPTIMIZATION_FLAGS:STRING="-march=native"
  -DANTs_CXX_OPTIMIZATION_FLAGS:STRING="-march=native"
  -DANTs_C_WARNING_FLAGS:STRING="-Wall -Werror"
  -DANTs_CXX_WARNING_FLAGS:STRING="-Wall -Werror"

### DIFF
--- a/CMake/ITKSetStandardCompilerFlags.cmake
+++ b/CMake/ITKSetStandardCompilerFlags.cmake
@@ -21,13 +21,18 @@
 
 include(ITK_CheckCCompilerFlag)
 include(ITK_CheckCXXCompilerFlag)
+if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.14.0)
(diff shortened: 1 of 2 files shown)


ntustison:
Thanks @hjmjohnson

PR #765, opened 2019-05-02
Switch generic label interpolator to now-merged ITK version

gdevenyi:
Since ITK GenericLabel now exists, remove the ANTs local copy.


ntustison:
@gdevenyi Thanks for this.

PR #768, opened 2019-05-18
Update to itkv5 api changes

hjmjohnson:


hjmjohnson:
@ntustison I rebased and merged these changes.  They were a bottleneck in building ANTs with newer ITK versions.  Recent changes removing the GenericLabelMapInterpolator from ANTs broke clean builds.  I'll work hard tomorrow if there are issues that need further work with this PR, but it has cleanly built on a local system.

ntustison:
Perfect.  Thanks a bunch @hjmjohnson .  The changes look straightforward.

gdevenyi:
@hjmjohnson GenericLabel was working fine for me when I pushed my change, did this problem crop up on newer ITK commits than the version that was the superbuild at the time?

Issue #776, opened 2019-06-12
Linux compilation failing on Travis


cookpa:
Travis has an older GCC that won't compile ANTs recently

https://travis-ci.org/ANTsX/ANTs/jobs/534217137

I managed to build successfully on Centos 6 with gcc version 7.3.1.

ntustison:
Thanks @cookpa .  So what do we do?  Can we remove this problematic travis build?

cookpa:
It might be possible to update the Travis GCC version, I'll have a go with GCC 5.

cookpa:
OK I can build on GCC 5 with Xenial. I tried to update the travis config but it got confused and ran on my merge commit and not my actual changes. 

Basically I am trying to get it to use Ubuntu Xenial, which has GCC 5 by default. I'm not sure it will work without explicitly setting up a matrix config.

cookpa:
OK it's running now, I will merge to master if successful.

cookpa:
Alas it still seems to run trusty even when I attempt to set Xenial. I am probably doing something wrong but I can't figure it out

 https://travis-ci.org/ANTsX/ANTs/jobs/545371818/config

cookpa:
@gdevenyi Thanks for all your work on this issue. It looks like the Linux build is now good, but there are some timeout failures on the Mac.

https://travis-ci.org/ANTsX/ANTs/builds/548722675?utm_source=github_status&utm_medium=notification

gdevenyi:
@cookpa I see that. It looks like the master branch travis didn't inherit my cache priming from the PR, in fact, it looks like some builds aren't getting cached at all based on the recent commits on master.

Sadly travis seems to be super flaky, see the http errors breaking some of the builds on this commit: https://travis-ci.org/ANTsX/ANTs/builds/549759369.

My next step here is to pare-down the ITK config so that it only builds the things that ANTs actually uses, this should alleviate the timeout issues.

cookpa:
Going to close this, thanks @gdevenyi for your work here

  https://github.com/ANTsX/ANTs/pull/792

and elsewhere to vastly improve the Travis environment for ANTs.

PR #779, opened 2019-06-13
Rewrite Travis Matrix

gdevenyi:
Saw #776 

I took at look at the canonical way to use all the different compilers and I'm writing a build matrix to try all the different versions.

WIP, missing OSX still as well as some newer llvm versions.


cookpa:
Thanks for the help @gdevenyi 

I'm not sure it's working correctly. You have got xenial running, which is better than I managed, but it seems the GCC version is not respecting the environment variables in all cases. For example:

https://travis-ci.org/ANTsX/ANTs/jobs/545419507#L313-L314

But then the GCC 5 job looks right

https://travis-ci.org/ANTsX/ANTs/jobs/545419509#L334-L335

gdevenyi:
Yup! Check again, I fixed the eval with an export

gdevenyi:
The PR builds will lag behind my personal travis because I keep canceling and updating when I see something not quite right, here's the exports correct:
https://travis-ci.org/gdevenyi/ANTs/builds/545424852

I'm not sure clang will work yet, will see.

gdevenyi:
Keep an eye here https://travis-ci.org/gdevenyi/ANTs/branches as I iterate on getting everything working.

TODO:
- osx clang
- osx gcc
- osx multiple versions
- osx ccache

cookpa:
Great it looks like all the compilers are correct now. But a bunch of builds are timing out. Not sure what to do about this. Maybe turn all tests off?

gdevenyi:
It looks like they're failing simply due to the compile taking too long.

If you compare, some of the tests on ANTsX are succeeding (I think) because the ccache is warm for the builds: https://travis-ci.org/ANTsX/ANTs/builds/545736154

I think what I'm going to do is change the build command to just do a "make ITKv5" and exit so that the cache is warmed with the ITK codes, then switch it back to see if that helps.

I'm testing my cmake upgrade is OSX first so I'll wait for those builds to completely finish so I can see the complete status of everything before another change.

gdevenyi:
Alright, so I think I've fixed all the configurations, and I discovered that ccache wasn't getting properly used in the entire superbuild, so this should fix some of the build time.

I accidentally forgot a file the first force push, so if you could cancel my oldest travis build the newer one is the one to test.

gdevenyi:
Last bug is caused by this:
https://github.com/Homebrew/homebrew-core/issues/41045

I can work around it, but I've requested it be fixed instead.

ntustison:
@gdevenyi Thanks for this.

gdevenyi:
Hurrah we're all ready to go. No more build failures, just timeouts.

The only thing left to do is prime the cache, so I'll swap the build script to just build ITK, and then swap it back, then its ready for merge.

gdevenyi:
Lets see if I fly too close to the sun.

ITKv5 cache priming is almost complete.

I've enabled VTK and adding a cache priming stage, and I've also force pushed the final travis config. Let them all run in order and at the end we should have working Travis across OSX and Linux.

cookpa:
@gdevenyi it looks like it failed looking for OpenGL. 

```
CMake Error at /usr/local/cmake-3.12.4/share/cmake-3.12/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find OpenGL (missing: OPENGL_opengl_LIBRARY OPENGL_glx_LIBRARY
  OPENGL_INCLUDE_DIR)
```
From what I can gather you need to install libgl1-mesa-dev, and possibly set  

  OPENGL_INCLUDE_DIR=/usr/include/GL

gdevenyi:
Yup! Come back from appointment to find this :)

gdevenyi:
Okay, *latest* build is the VTK priming with added packages. You can cancel the other builds to get it going sooner.

cookpa:
I think it updates itself, I don't see the old build any more. It's got past the old failure point so let's see if this will be the one.

gdevenyi:
OSX has some broken libraries right now which prevent VTK from building. For now, VTK remains disabled in travis builds.

Re-priming ITKv5 cache because I did a dumb and didn't include both VTK and ITK in the last round, then final commit.

gdevenyi:
Looks like Travis got "stuck", can someone with rights poke (cancel/restart) the stuck jobs? https://travis-ci.org/ANTsX/ANTs/builds/547733369

cookpa:
@gdevenyi the Linux jobs all seem to exit after building ITK. I'll let you take a look before I cancel the build

gdevenyi:
Yes, that's correct, I'm priming the ccache so that all the builds will finish completely, so that means I need to "stop early" so I get a successful build and a cache store. Once these succeed I have one more force-push to re-enable the full build and the tests, and we'll be good to go.

cookpa:
OK restarted

gdevenyi:
We're good to go for merge.

The only failure was due to an HTTP failure to grab the test data during the ANTs portion of the build.

If you want a perfectly clean build before more, hit restart here: https://travis-ci.org/ANTsX/ANTs/jobs/547998872

ntustison:
Thanks @gdevenyi .  We'll go ahead and merge and then address any issues, if they arise.

PR #782, opened 2019-06-20
Not install binaries switch

muschellij2:
Is it possible to set a switch or CMake flag to not install the binaries?  These are installed when installing on ANTsRCore and they bloat the package size, but are not used as the .so and .a files are the ones linked to.  I don't know about https://github.com/ANTsX/ANTsPy, but it would make sense to have a flag.  I think this could also help with the windows build @adigherman 
@1alebra2


muschellij2:
I'm not really recommending a PR here @stnava, but just showing that we can turn these off and still get a full install on ANTsRCore

1alebra2:
As far as I know it is only possible to install the binaries in a separate
directory using the r cmd install routine. About devtools I am not sure. I hope
this helps?

> Am 20.06.2019 um 17:45 schrieb John Muschelli <notifications@github.com>:
> 
> Is it possible to set a switch or CMake flag to not install the binaries?
These are installed when installing on ANTsRCore and they bloat the package
size, but are not used as the .so and .a files are the ones linked to. I don't
know about https://github.com/ANTsX/ANTsPy, but it would make sense to have a
flag. I think this could also help with the windows build @adigherman
> @1alebra2
> 
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/ANTsX/ANTs/pull/782
> 
> Commit Summary
> 
> trying static for antsregistration
> Merging with upstream
> Merged with upstream
> Merging with ANTs
> I think there's a problem with GaussianListSampleFunction king branch
'upstream/master'

muschellij2:
These are installed by default in the R installation the way it's set up with the CMake and configure file.  I'm trying to stop that because they are not needed but simply the lib files.

gdevenyi:
A few things, here, you've changed the ITK source to your own repo, and you have a huge number of commits.

Can you please remove the changes to the ITK source, and squash this into a single commit?

cookpa:
Canceled the Travis build, as it has a bunch of failures. It would be nice to have a switch that just installs the libs, as well as a switch that just installs the binaries. People building static executables don't need the libs, right?

gdevenyi:
@cookpa that is correct, I had configured it that way during my cleanup/fix of the build system, however @stnava reverted the library install change here https://github.com/ANTsX/ANTs/commit/b1d9bf8cdf6558c4499aa2ca1f1b39188e06e2d1 because of it breaking hows ANTsR builds.

I can try again at adding a config option to turn on/off the library install so that ANTsR can use that.

gdevenyi:
This PR has a bunch of uglyness because its a master->master merge. If we could get this on a seperate branch it would keep the history clean.

gdevenyi:
Could someone please close this as it is made using @muschellij2's working branch and triggers a Travis build every time they work on it.

To evaluate this change properly it needs to be on a separate branch with just the proposed changes.

Please also cancel the Travis builds as it is holding up PR #973

PR #786, opened 2019-06-25
Remove depreciated OSX version from travis config

gdevenyi:
As per the note here: https://travis-ci.org/ANTsX/ANTs/jobs/549759403

Travis is removing xcode6.4 in "January 2019"


gdevenyi:
As usual, Travis timeouts but I just removed a matrix elements so there's nothing else wrong.

cookpa:
Thanks, merging this.

PR #787, opened 2019-06-25
Convert cmake to use ninja build system in travis for improved build performance

gdevenyi:
Hoping this may fix the timeouts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ matrix:
 #          sources:
 #            - ubuntu-toolchain-r-test
 #          packages:
+#            - ninja-build
 #            - cmake
 #            - cmake-data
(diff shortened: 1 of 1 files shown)


gdevenyi:
Hurrah, it looks like switching in Ninja saves ~20 minutes of build time. Hopefully this will be enough to rescue all the builds that are just over the limit.

gdevenyi:
Again, the only failures we see here are travis connectivity issues failing to download the ExternalData, builds and tests pass, and are faster.

gdevenyi:
We're all good for a merge here. My next project will be figuring out how to cache the ExternalData components so that these random HTTP errors stop breaking the build.

ntustison:
@gdevenyi Thanks for this.

PR #790, opened 2019-06-26
WIP: Trim back ITK and VTK superbuild using WhatModules{ITK,VTK}.py

### LINKED ISSUE #789: Proposal: Minimal VTK/ITK superbuild configs
Asking around on the ITK discourse, there are scripts for both VTK and ITK that will tell you what modules a package uses.
https://discourse.itk.org/t/trimming-down-itk-build-to-just-used-parts/1995

I'm wondering whether its worth it to shrink the ITK/VTK builds inside ANTs to just what is being used.

Is anyone else building against these internal versions?

gdevenyi:
First try at shrinking the ITK and VTK superbuilds.

Not tested at all yet.

Fixes #789


gdevenyi:
So far this is looking okay. The failures are just the regular Travis being flaky, ctest passes.

I need to test with a local build more extensively to make sure we don't lose any IO modules or such.

gdevenyi:
Any thoughts on if this effort is worthwhile?

ntustison:
Thanks for this @gdevenyi .  It seems as though there are a number of proposed improvements.  Can you help me sort through what needs to be approved and in what order?

gdevenyi:
@ntustison the idea here is to shrink the superbuild to the minimal components of vtk/itk that are actually used. I started by disabling the default modules and then used the python scripts provided by ITK/Vtk which I learned from https://discourse.itk.org/t/trimming-down-itk-build-to-just-used-parts/1995/2 to re add the used components.

Is this effort to slim down the build worthwhile?

ntustison:
@gdevenyi I would guess that that would be useful but if it's a hassle, I don't think that I would prioritize it.

gdevenyi:
Can someone please kick this build now that the cache is warm, I'd like to see how it compares build time wise to the mainline. Thanks.

ntustison:
Sure, just did.

gdevenyi:
Doesn't seem to save any time/space.

PR #792, opened 2019-06-27
WIP: Further Travis/Build Improvements

gdevenyi:
It looks like (on Travis at least) that the ccache symlinks interfere with the cmake ccache integration added when I rewrote the travis matrix.

This is a test to see if caching works properly when we explicitly ignore the ccache symlinks dir.


gdevenyi:
This refers specifically to the travis linux GCC builds, where ccache automatically adds symlinks, these builds still take 1 hour + when the rest don't.

gdevenyi:
Discovered some more improvements, replicating less of the travis config now.

Also, found out that we weren't using the gold linker due to incomplete cmake configuration stolen from ITK. I've updated that.

gdevenyi:
*hopefully* figured out caching externaldata so that http will stop failing.

gdevenyi:
Added caching of homebrew downloads since that's also flaky....

gdevenyi:
And we're all green!
- cleaner build matrix with less replication
- caching of *everything* I could manage
- update to use cmake's externalobject module
- update to newer version of ITKSetFlag cmake file
- fix a build bug with newer cmake versions not finding ITK because its 5.1 now

stnava:
That is a really tremendous effort @gdevenyi  .... many thank yous.

stnava:
I do think it would be great to have a minimal ITK/ANTs build, if possible.   I don't have an estimate of how much time/memory this would save but it would ( potentially ) facilitate using our resources in the context of lambda functions / serverless applications.

ntustison:
@gdevenyi Yes, thank you very much for this.

gdevenyi:
I am supremely frustrated that after all that my caches don't transfer from the PR into master. I will continue my work in #790 to slim the build in the hopes things will eventually build, otherwise, we need to go through a cache-priming cycle of just building ITK, and then switching back to the current configuration so the build succeeds.

PR #820, opened 2019-08-18
COMP: Allow building with ITK_LEGACY_REMOVE turned on

hjmjohnson:
The VectorExpandImage filter specialization is no longer needed because
the ExpandImage filter now works with both vector and scalar types
uniformly.


ntustison:
Thanks @hjmjohnson

Issue #828, opened 2019-09-03
How to remove the negative region in -BSpline interpolation


John1231983:
I have a template image and a moving image. I want to register the moving image to the template image using `BSpline` interpolation. It worked fine but the problem is that how to remove the negative region in moved image (results by `BSpline`)

I did it by using a mask that results in the threshold of moved image. However, I do not know which is threshold value should we choose. Currently, I am using 0.5 (just by visualize the mask). Could you suggest any way to select the threshold value? or the better way to remove the negative value in `BSpline` interpolation.

This is the source file
[OASIS_template0.nii.gz](https://github.com/ANTsX/ANTs/files/3569753/OASIS_template0.nii.gz)
[OASIS-TRT-20-1_brain.nii.gz](https://github.com/ANTsX/ANTs/files/3569754/OASIS-TRT-20-1_brain.nii.gz)
```
#Step 1
$ANTSPATH/antsRegistrationSyN.sh  -d 3 -f OASIS_template0.nii.gz -m OASIS-TRT-20-1_brain.nii.gz -o OASIS-TRT-20-1_brain_reg.nii.gz -n 16 -t a
#Step 2
$ANTSPATH/antsApplyTransforms -d 3 -i OASIS-TRT-20-1_brain.nii.gz -r OASIS_template0.nii.gz -o OASIS-TRT-20-1_brain_reg.nii.gz -n BSpline -t OASIS-TRT-20-1_brain_reg.nii.gz0GenericAffine.mat
#Step 3
ThresholdImage 3 OASIS-TRT-20-1_brain_reg.nii.gz OASIS-TRT-20-1_brain_reg_mask.nii.gz 0.5 Inf 1 0
#Step 4
ImageMath 3 OASIS-TRT-20-1_brain_reg_remove_negative.nii.gz m OASIS-TRT-20-1_brain_reg_mask.nii.gz OASIS-TRT-20-1_brain_reg.nii.gz
```

cookpa:
If you want to remove negative values:

```
ThresholdImage 3 input.nii.gz mask.nii.gz 0 Inf
ImageMath 3 nonNegative.nii.gz m input.nii.gz mask.nii.gz
```

Negative values are now set to 0.

John1231983:
I have tried to set it 0 but the mask looks not good. Also, it may keep non-zero value in the background of moved image (results in multiply mask and bspline result)
[...]

PR #837, opened 2019-09-09
Fix ThresholdImage

gdevenyi:
Fixes bug noted at #828

### DIFF
--- a/Examples/ThresholdImage.cxx
+++ b/Examples/ThresholdImage.cxx
@@ -73,6 +73,7 @@ typename TImage::Pointer BinaryThreshold_AltInsideOutside_threashold(
   typedef itk::BinaryThresholdImageFilter<TImage, TImage> InputThresholderType;
   typename InputThresholderType::Pointer inputThresholder =
     InputThresholderType::New();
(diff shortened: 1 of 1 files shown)


cookpa:
Looks good thank you

ntustison:
Thanks @gdevenyi .  However, just saw [this](https://github.com/ANTsX/ANTs/pull/837/commits/0f8d4ac7a923f263d5bf20e1f9675b33370ef30e#diff-5003d0db4237942a8abedf7385c00837L81-L84) which isn't obvious behavior and bypasses the [internal check](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/Thresholding/include/itkThresholdImageFilter.hxx#L80-L84) in the filter.  Unless there are good reasons for including this, I would vote to remove it.

gdevenyi:
Okay, will remove completely, I was simply maintaining the existing setup, additional PR incoming.

gdevenyi:
PR #839

ntustison:
I figured.  But thanks for fixing it.

Issue #835, opened 2019-09-07
Fix gcc<5 builds or error out early


gdevenyi:
We're getting regular reports of compile failure due to old gcc, we should fix or error out in cmake early.

cookpa:
On my to-do list for today

gdevenyi:
Poke me and I'll be happy to re-add older compilers to the travis config.

cookpa:
This gives an error if you try to use GCC < 5. 

https://github.com/ANTsX/ANTs/tree/minGCC

But I wonder if we can fix it, I don't think GCC > 5 is officially an ITK requirement yet.

I'm going to try to build ITK independently on an older GCC.

cookpa:
Actually looking at the older failure logs with GCC 4, the ITK build completes normally, but then ANTs fails when linking against it.

cookpa:
I get the same error if I build ITK offline and use it as system ITK for ANTs. I don't have the time or expertise to get more into this, so I suggest we will have to start requiring GCC 5

Issue #834, opened 2019-09-07
Build error with the current branch


muratmaga:
Hi,
I am trying to build ANTs on my new laptop through wsl and I am getting the error below. It doesn't appear to be a case of missing libraries or along those lines...

[root@Maga-XPS15 ANTs-build]# make
[ 47%] Built target ITKv5
[ 52%] Performing forcebuild step for 'ANTS'
[ 58%] Performing build step for 'ANTS'
[  0%] Building CXX object Examples/CMakeFiles/antsUtilities.dir/antsRegistrationTemplateHeader.cxx.o
In file included from /root/ANTs/Examples/antsRegistrationTemplateHeader.h:21:0,
                 from /root/ANTs/Examples/antsRegistrationTemplateHeader.cxx:1:
/root/ANTs-build/ITKv5-install/include/ITK-5.1/itkLabelImageGenericInterpolateImageFunction.h:40:70: error: expected ‘class’ before ‘TInterpolator’
 template <typename TInputImage,template<typename, typename> typename TInterpolator, typename TCoordRep=double >
                                                                      ^
In file included from /root/ANTs-build/ITKv5-install/include/ITK-5.1/itkLabelImageGenericInterpolateImageFunction.h:127:0,
                 from /root/ANTs/Examples/antsRegistrationTemplateHeader.h:21,
                 from /root/ANTs/Examples/antsRegistrationTemplateHeader.cxx:1:
/root/ANTs-build/ITKv5-install/include/ITK-5.1/itkLabelImageGenericInterpolateImageFunction.hxx:27:70: error: expected ‘class’ before ‘TInterpolator’
 template<typename TInputImage, template<typename, typename> typename TInterpolator , typename TCoordRep>
                                                                      ^
/root/ANTs-build/ITKv5-install/include/ITK-5.1/itkLabelImageGenericInterpolateImageFunction.hxx:58:70: error: expected ‘class’ before ‘TInterpolator’

 template<typename TInputImage, template<typename, typename> typename TInterpolator , typename TCoordRep>
                                                                      ^
make[5]: *** [Examples/CMakeFiles/antsUtilities.dir/antsRegistrationTemplateHeader.cxx.o] Error 1
make[4]: *** [Examples/CMakeFiles/antsUtilities.dir/all] Error 2
make[3]: *** [all] Error 2
make[2]: *** [ANTS-prefix/src/ANTS-stamp/ANTS-build] Error 2
[...]

Issue #832, opened 2019-09-06
Any post 2.3.1 release coming?


yarikoptic:
Here I am again inquiring. 
I have tried to build 2.3.1 a bit back on debian sid to get some not so clear to me "why" error:
```
/usr/bin/c++    -g -O2 -fdebug-prefix-map=/build/ants-2.3.1=. -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2  -Wall -Wcast-align -Wdisabled-optimization -Wextra -Wformat=2 -Winvalid-pch -Wno-format-nonliteral -Wpointer-arith -Wshadow -Wunused -Wwrite-strings -funit-at-a-time -Wno-strict-overflow -Wno-deprecated -Wno-invalid-offsetof -Woverloaded-virtual -Wstrict-null-sentinel -DCXX_HAS_DISABLE_OPTIMIZATION_FLAG   -o CMakeFiles/cmTC_1f1dd.dir/src.cxx.o -c /build/ants-2.3.1/obj-x86_64-linux-gnu/CMakeFiles/CMakeTmp/src.cxx
/build/ants-2.3.1/obj-x86_64-linux-gnu/CMakeFiles/CMakeTmp/src.cxx:1:1: error: expected unqualified-id before '-' token
 -O0
```
so before digging in or testing current snapshot thought to check may be a new release coming since there was a good amount of development (currently you are at v2.3.1-256-g03d751a4) since then

gdevenyi:
Hi, to provide any help with builds you'll have to provide a bunch more details.

As far as we're aware, there's no issues with gcc 5+ and clang-3.6+. Perhaps you need to check your compiler version.

gdevenyi:
As for releases, I'm working on it when I have time, discussion here: https://github.com/ANTsX/ANTs/issues/802

yarikoptic:
Thanks for the reply. The question is either we should try to troubleshoot build of 2.3.1 (all the versions as of Debian buster and sid are of interest for me, gcc 9 on sid,  and I can provide detailed logs later) or should I wait and aim for the next one and this try to build current master snapshot

gdevenyi:
Full logs are always appreciated, if we don't build in some configurations, we should figure out why.

yarikoptic:
[...]

Issue #802, opened 2019-07-12
Packaging and releases


gdevenyi:
With the streamlining of the travis configuration, it is possible to have travis push a rendered file onto the github releases page after a successful build.

That, in combination with fixing the cmake configuration so that -DCMAKE_PREFIX_PATH is honoured should allow us to use cpack to create TGZ statically linked tarballs for a variety of distribution, or even cpack rpms or debs to install to say, /opt/ANTs/major.minor.patch

This will require a bit more re-engineering of the cmake config, and a bit more travis work, the question is, is it considered a reasonable goal by the devs?

ntustison:
@gdevenyi Although @stnava should weigh in, I think that would be great.

gdevenyi:
Things to think about for regular releases:
https://semver.org/
https://cmake.org/cmake/help/latest/module/CPack.html
https://docs.travis-ci.com/user/deployment/releases/

stnava:
regular, automated releases would certainly make many people happier,
including me.

brian

justbennet:
This seems like it would be a desirable thing.  I am only the system administrator, so my only comment would be that it would be great if the resulting binary distributions either worked on variety of systems or there were versions for a variety of systems.  I have, right now, Ubuntu 16.04, 18.04, and CentOS 7.  CentOS 8 will be added to the fold by the end of September.

I did notice today that there aren't binary releases for any of 2.3.1, 2.3.0, or 2.2.0.  Last binary release seems to be 2.1.0.

I am getting compilation errors from ITK 5 when following the instructions at https://github.com/ANTsX/ANTs/wiki/Compiling-ANTs-on-Linux-and-Mac-OS, and I'd be quite happy to have a binary distribution available.

cookpa:
Got some releases set to build in #1393. It's a default superbuild, but I'm asking for feedback on potential options in https://github.com/ANTsX/ANTs/discussions/1397

cookpa:
Some sub-issues (like semantic versioning) would be good to revisit at some point but I hope that the progress to date has increased happiness https://github.com/ANTsX/ANTs/releases

PR #841, opened 2019-09-09
Update CMAKE to build static and shared builds properly, fix "make install"

gdevenyi:
Upgrades to the build system in cmake and travis
- update the default flag code from ITK upstream
- Error out for gcc<5 c/o @cookpa 
- Strip out hacky versioning bits to use the lovely git-tag driven versioning
- Fix travis clone so that it gets all tags
- Fix cmake scripts to properly honour CMAKE_INSTALL_PREFIX so that "make install" works
- Set PIC/PIE using canonical cmake method
- Remove hackery for installing scripts to use "make install" infrastructure instead
- use travis-command-wrapper.py for nice travis timeout wrapping

Todo:
- [x] don't install libraries if static build
- [x] don't cpack libraries if static build
- [x] install vtk/itk if shared build
- [x] cpack vtk/itk if shared build
- [ ] ants-config.sh script for setting up environment after https://github.com/BIC-MNI/minc-toolkit-v2/blob/master/minc-toolkit-config.unix.sh.cmake
- [x] audit "make install" for scripts


gdevenyi:
This is all in prep for making releases on the github release page automatically.

Thing that remains is that when I cpack up a superbuild, I get version 0.1.1 and I can't figure out where thats coming from ... yet.

gdevenyi:
looks like cpack works properly inside ANTs-build directory, which is okay.

I'm going to try to figure out how to skip packaging the libs if a static build is done since they're not needed.

gdevenyi:
Still needed
- what optimisation settings should we use for automated release builds
- should we include debug symbols
- are all scripts listed in Script install list
- what kind of cpack packages would we like to publish
- don't package libraries if static build
- properly install ITK/VTK if shared build
- what compiler version should we use for the official release binaries

gdevenyi:
Fixes #835 #832 #834 #802 #612 #443

gdevenyi:
Needs another kill of existing build if anyone is around

ntustison:
Okay, that should do it.

gdevenyi:
Can someone please kill the running build in favour of the new one waiting? Thanks!

ntustison:
Sure.

gdevenyi:
P.S. If you go to https://travis-ci.org/ANTsX/ANTs/settings there is an "Auto Cancellation" setting, where superseding PR pushes will cancel older ones automatically.

ntustison:
It looks like it's already on.

Screen Shot 2019-09-10 at 9 58 45 AM

gdevenyi:
Ah, more Travis bugs no doubt :-)

cookpa:
Well, it does say "Builds will only be canceled if they are waiting to run, allowing for any running jobs to finish."

gdevenyi:
Ah good catch @cookpa

gdevenyi:
Again if someone wouldn't mind, cancel all but the latest? Thanks :+1:

cookpa:
Done

gdevenyi:
Got it backwards :)
Cancel: https://travis-ci.org/ANTsX/ANTs/builds/583255037
Restart: https://travis-ci.org/ANTsX/ANTs/builds/583258441

cookpa:
OK sorry

gdevenyi:
Morning, could someone please cancel all but the most recent build. Thanks :)

ntustison:
Sure.

cookpa:
@gdevenyi I saw your other closed PR about the gold linker. When adding the GCC version requirement for the ITK build, I found code that checked for the gold linker here

https://github.com/ANTsX/ANTs/blob/47432e61f6a369b3661eb0b4416607f8c8fcb9fa/CMake/ITKSetStandardCompilerFlags.cmake#L274

I removed code that set special conditions for GCC 4.X, which we no longer support. But I don't think you made those same changes when you superseded my PR, so it looks like gold is still not used for the ITK build.

gdevenyi:
Thanks for following up @cookpa , the ITKSetStandardCompilerFlags.cmake is a copy-and-paste from ITK upstream, with a search and replace of ITK with ${PROJECT_NAME} so we can use it in the superbuild.

To my reading, these lines test if gold works, and enable it if so:
https://github.com/gdevenyi/ANTs/blob/cmake-improvements/CMake/ITKSetStandardCompilerFlags.cmake#L269-L274

If you take a look at the current PR build logs, https://travis-ci.org/ANTsX/ANTs/jobs/583679172, and Ctrl-F "have_gold" we see:
```
-- Performing Test have_gold
-- Performing Test have_gold - Success
```

I also see the ``-fuse-ld=gold`` flag being set:
https://travis-ci.org/ANTsX/ANTs/jobs/583679172#L9047

Am I missing something?

cookpa:
No, upon closer inspection I think you are right. Because GCC can no longer be less than 5, the gold failure conditions cannot be true now.

Sent from my iPhone

On Sep 11, 2019, at 12:15 PM, Gabriel A. Devenyi <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks for following up @cookpa<https://github.com/cookpa> , the ITKSetStandardCompilerFlags.cmake is a copy-and-paste from ITK upstream, with a search and replace of ITK with ${PROJECT_NAME} so we can use it in the superbuild.

To my reading, these lines test if gold works, and enable it if so:
https://github.com/gdevenyi/ANTs/blob/cmake-improvements/CMake/ITKSetStandardCompilerFlags.cmake#L269-L274

If you take a look at the current PR build logs, https://travis-ci.org/ANTsX/ANTs/jobs/583679172, and Ctrl-F "have_gold" we see:

-- Performing Test have_gold
-- Performing Test have_gold - Success


I also see the -fuse-ld=gold flag being set:
https://travis-ci.org/ANTsX/ANTs/jobs/583679172#L9047

Am I missing something?

gdevenyi:
I've got a lot of changes here, can someone take a look and comment?

Next up is adding the code so that ITK/VTK is installed/cpacked if a SHARED build is made (and not SYSTEM)

cookpa:
@gdevenyi I compiled from the head of your `gdevenyi:cmake-improvements` branch and it compiles with no errors, but I don't see the install directory anywhere. Am I missing something?

gdevenyi:
@cookpa the install target is within ANTS-build, not in the upper level superbuild.

cookpa:
Great, and it copies the script too. Nice!

gdevenyi:
As per comment 1, I still need to add some scripting to install the VTK/ITK libs *if* its a shared build and *if* external stuff is not used.

After that, I need these questions answered:
https://github.com/ANTsX/ANTs/pull/841#issuecomment-529726864

to setup automatic deployment to github releases.

gdevenyi:
Hi, I've run into a bit of a roadblock for the shared build. In this commit: https://github.com/ANTsX/ANTs/pull/841/commits/b45ad316da7a84fd4deaac95fb32da8eacc4af16 I fix the building with shared libraries, however the install command here:
https://github.com/gdevenyi/ANTs/blob/cmake-improvements/Examples/CMakeLists.txt#L55-L62

doesn't result in the ANTS libraries being installed during make install. It looks "right" to me, but cmake is still a bit opaque to me. Everything works as expected if I copy manually and set the LD_LIBRARY_PATH so I know I got the build right, but the install isn't happening.

Can anyone assist?

ntustison:
@hjmjohnson would be the go-to person for this.

gdevenyi:
Hurrah, a green build across the board.

ntustison:
@gdevenyi Great.  So is this ready to merge?

gdevenyi:
@ntustison I'd like to fix the install of the ANTs libraries during shared build as per https://github.com/ANTsX/ANTs/pull/841#issuecomment-533822621 

I'll need to do a bit more experimenting before that.

gdevenyi:
Paging @hjmjohnson, I'm having trouble with "make install" not installing libraries from this snip of code:
https://github.com/gdevenyi/ANTs/blob/cmake-improvements/Examples/CMakeLists.txt#L55-L62

Can you offer any assistance?

gdevenyi:
Success! Somehow I mixed up CMAKE_BUILD_SHARED and BUILD_SHARED_LIBS so the static build macro was always running, hence the proper install wasn't running. Fixes incoming.

I can't help but feel make cmake needs a "strict mode" so I get an error instead of if(CMAKE_BUILD_SHARED) returning FALSE because its non existent.

gdevenyi:
Someone can cancel: https://travis-ci.org/ANTsX/ANTs/builds/590568781

After a green Travis pass this is ready for merge.

This PR fixes "make install" from the ANTs-build directory. It will install libraries if shared, not if static build, including proper logic for the superbuild ITK/VTK. It uses the RPATH feature of linking so that LD_LIBRARY_PATH doesn't need to be adjusted, avoiding conflicts with system libraries. CPack now works as well, which means my next PR will be to add automatic generation of CPACK installation files for tagged releases of ANTs.

cookpa:
> Someone can cancel: https://travis-ci.org/ANTsX/ANTs/builds/590568781

Canceled

gdevenyi:
Travis hiccup, please restart: https://travis-ci.org/ANTsX/ANTs/jobs/590569678

gdevenyi:
All green and ready for merge.

gdevenyi:
I suspect master will need a restart of the build process after at least one failed build to bootstrap the cache.

ntustison:
@gdevenyi Thanks for this.

gdevenyi:
Master could use a restart now to take advantage of the cache now

ntustison:
Quick question---just did a fresh install and now the binaries install to ``${BINARY_DIR}/ANTs-build/Examples/`` which is a change in behavior.  Is this what we want?

gdevenyi:
@ntustison This PR updated the CMAKE so that the build properly honours CMAKE_INSTALL_PREFIX.

The install process is "cmake -DCMAKE_INSTALL_PREFIX=/path/to/install/to; make; cd ANTS-build; make install``

This is to make ANTs conform to how most other software is built and installed (automake, ./configure --prefix=/path/to/install/to; make; make install).

The documentation was already updated: https://github.com/ANTsX/ANTs/wiki/Compiling-ANTs-on-Linux-and-Mac-OS

ntustison:
Understood.  But shouldn't the default be to keep the existing behavior?

cookpa:
The existing behavior was to ignore the install prefix, put bin/ inside the build directory and require users to manually copy the scripts, which was not great.

I guess you could make the default prefix ${buildDir}, to get closer to the old behavior.

I fear the "cd ANTs-build" bit is going to become an FAQ. Is there any way to make it work from the superbuild directory, ie the same place that make is called from? Or if not, to print a message telling the user to cd to ANTs-build?

gdevenyi:
> I guess you could make the default prefix ${buildDir}, to get closer to the old behavior.

Just tried that, but with all the manual hackery removed it doesn't work:
```
CMake Error at CMake/PreventInBuildInstalls.cmake:5 (message):
  The current CMAKE_INSTALL_PREFIX points at the build tree:

    /scratch/antsbuild

  This is not supported.
Call Stack (most recent call first):
  Common.cmake:15 (include)
  SuperBuild.cmake:15 (include)
  CMakeLists.txt:67 (include)


-- Configuring incomplete, errors occurred!
```

> I fear the "cd ANTs-build" bit is going to become an FAQ. Is there any way to make it work from the superbuild directory, ie the same place that make is called from? Or if not, to print a message telling the user to cd to ANTs-build?

I agree, I haven't figured that out yet, but I felt this PR was "big enough" to achieve the goal of having a functional STATIC and SHARED build, with working travis and tests. I was hoping a cmake expert like @hjmjohnson might provide an assist eventually..

cookpa:
For my 2c, having to build in ever more diverse environments, VMs and containers, I think a proper make install is a very good thing. Thanks for all your work making it happen.

ntustison:
Agreed.  This is just a rather significant change and so I just want to make sure that we're minimizing the resulting difficulties for us and the users.

PR #844, opened 2019-09-14
COMP: Updated `enum` to `enum class` for c++11

hjmjohnson:
Improve use of enumerations to new features in C++11.  A set of
refactorings to move from "c enumerations" to "strongly typed
enumerations" is included.

See: http://www.stroustrup.com/C++11FAQ.html#enum

The ITKSmoothing module now has a library that must be linked, so the
module dependancies must be moved from COMPILE_DEPENDS to DEPENDS.

Proper ENABLE_SHARED, and ITK{MODULENAME}_EXPORT with
`#include "ITK{MODULENAME}_Export.h"` instrumentation for
the ostream operators used with the enumerations are added.

### DIFF
--- a/Examples/ANTS.cxx
+++ b/Examples/ANTS.cxx
@@ -190,14 +190,14 @@ private:
   if( argc == 3 && ( std::stoi( argv[1] ) != '-' || std::stoi( argv[1] ) != 2 || std::stoi( argv[1] ) != 3 ) )
     {
     itk::ImageIOBase::Pointer fixedImageIO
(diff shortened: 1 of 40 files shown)


hjmjohnson:
The one failing case on gcc8 still passes all tests:
```
100% tests passed, 0 tests failed out of 130
```

The other mac builds are taking a really long time to get started.  It looks like they may take a few days to complete.

ntustison:
Thanks @hjmjohnson

PR #861, opened 2019-10-01
Fix dynamic VTK linking for shared build

spam-depository:
Current code generates broken link.txt files with "BUILD_SHARED_LIBS=ON" for code using VTK modules (e.g. Examples/antsVol and Examples/antsSurf).

Tested with CMake 3.10.2 and 3.15.3.


stnava:
thank you

gdevenyi:
Oops, good catch!

PR #866, opened 2019-10-07
PERF:  change computation type for fft:  float --> double

ntustison:
Mirrors the N4 changes in ITK [here](https://github.com/InsightSoftwareConsortium/ITK/commit/389da5d8e5dbd433a506c750c7b914ead1482d54#diff-4fd7d6e8f77462fbcf32c75bf3bfe9f8)

### DIFF
--- a/Examples/N4BiasFieldCorrection.cxx
+++ b/Examples/N4BiasFieldCorrection.cxx
@@ -466,33 +466,13 @@ int N4( itk::ants::CommandLineParser *parser )
   if( outputOption && outputOption->GetNumberOfFunctions() )
     {
     /**
-                    * Reconstruct the bias field at full image resolution.  Divide
-                    * the original input image by the bias field to get the final
-                    * corrected image.
-                    */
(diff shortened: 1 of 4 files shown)


gdevenyi:
Huh, I think I've been bitten by this bug :)

ntustison:
Yeah, I should've updated sooner.

cookpa:
Does this address a known bug?


On Oct 7, 2019, at 9:37 PM, Nick Tustison <notifications@github.com<mailto:notifications@github.com>> wrote:


Yeah, I should've updated sooner.

ntustison:
No, not a known bug, but if you follow the link in my original posting, you'll see that @hjmjohnson encountered a float vs. double issue when processing many scans through N4.  Since N3 and N4 share the same histogram sharpening function, I thought it should be propagated to our ANTs version of N3.  The other commit was due to a recent change in the N4 code where they exposed the ``ReconstructBiasField`` function so I took advantage of that to trim down some lines of code in ``N4BiasFieldCorrection.cxx``.

PR #869, opened 2019-10-09
COMP:  GCC 9 enhanced warnings.

ntustison:
Round 1 for:

https://github.com/ANTsX/ANTs/issues/854

PR #871, opened 2019-10-10
[WIP] switch to using verbose make for travis to expose build problems clearly

gdevenyi:
Trying to address #869 issues with build, getting rid of ninja.

Need to see if this increases the build time too much.


gdevenyi:
Can someone please restart this and see if it'll work this time? :)

ntustison:
Sure.

gdevenyi:
My ninja solution works, so we'll stick with that.

PR #875, opened 2019-10-11
Switch to verbose Ninja properly

gdevenyi:
After a bunch of googling it seems this is the proper way to force ninja to be verbose. Tested locally seems to work

Hopefully this works on Travis as well.


gdevenyi:
Hurrah, verbose ninja travis build, now we can actually see why things fail.

Issue #872, opened 2019-10-10
Install libraries in /usr/lib64 instead of /usr/lib


TheChymera:
**Describe the bug**
I would like to build ANTs so that the libraries get installed into `usr/lib64` instead of `usr/lib` . I believe this should be configurable via the CMake build system.

**To Reproduce**
I am building ANTs with:

```
cmake -C /var/tmp/portage/sci-biology/ants-9999/work/ants-9999_build/gentoo_common_config.cmake -G Unix Makefiles -DCMAKE_INSTALL_PREFIX=/usr -DUSE_SYSTEM_ITK=ON -DITK_DIR=/usr/include/ITK-5.0/ -DBUILD_TESTING=OFF -DUSE_VTK=OFF -DUSE_SYSTEM_VTK=OFF -DCMAKE_BUILD_TYPE=Gentoo -DCMAKE_TOOLCHAIN_FILE=/var/tmp/portage/sci-biology/ants-9999/work/ants-9999_build/gentoo_toolchain.cmake  /var/tmp/portage/sci-biology/ants-9999/work/ants-9999
```

This is the full [build.log](https://github.com/ANTsX/ANTs/files/3714834/build.log).

gdevenyi:
PRs welcome: https://rix0r.nl/blog/2015/08/13/cmake-guide/

TheChymera:
@gdevenyi is the build 64-bit only?

TheChymera:
Similar to re-report in https://github.com/ANTsX/ANTs/issues/1061

Fixed by the pending PR: https://github.com/ANTsX/ANTs/commit/329a7be631b33294d9cf3d58e8cdeb62e0aa993c

cookpa:
Fixed in #1062

PR #867, opened 2019-10-08
Fix install step for SYSTEM/ITK/VTK, fix corner case in cmake config.

gdevenyi:
Addresses https://github.com/ANTsX/ANTs/issues/857#issuecomment-538966691

PR #877, opened 2019-10-12
Install into appropriate platform directories

TheChymera:
Via `GNUInstallDirs`:

* tested, producing: [ants_files.txt](https://github.com/ANTsX/ANTs/files/3721196/ants_files.txt)
* fixes https://github.com/ANTsX/ANTs/issues/872

@gdevenyi


gdevenyi:
Thanks @TheChymera this is a good start but it doesn't modify the superbuild codepath we had trouble with in my previous PR. That's the way most people are installing ANTs still.

I think this can be addressed if you could provide some detail on how you got your ITK install to honour this lib vs lib64 setting, since we'll need to do something like that to the ITK and VTK in the superbuild.

TheChymera:
>I think this can be addressed if you could provide some detail on how you got your ITK install to honour this lib vs lib64 setting, since we'll need to do something like that to the ITK and VTK in the superbuild.

This patch was all that was needed (in addition to your logic patch from https://github.com/ANTsX/ANTs/pull/867 --- though it seems unrelated) to successfully produce and install the attached file list. Does this seem unlikely, or is anything missing?

gdevenyi:
As always, the exact cmake configuration determines which things are built and which things are installed. Your file list does not provide suffcient information on whether this patch works. The exact cmake config is also required.

My guess is you are not testing the SuperBuild path since you are specifying USE_SYSTEM_ITK AND USE_SYSTEM_VTK. Most users are using the SuperBuild, which (in the BUILD_SHARED_LIBS config) triggers this codepath:
https://github.com/ANTsX/ANTs/blob/master/ANTS.cmake#L178

Notice that this still hard-codes a libs directory. Please test this codepath using the ANTs-documented installation instructions.

TheChymera:
@gdevenyi could you provide a link to the install instructions you are referring to, I found no unabbiguous resource in the README.

gdevenyi:
https://github.com/ANTsX/ANTs/wiki/Compiling-ANTs-on-Linux-and-Mac-OS

TheChymera:
@gdevenyi ok, so I gave it a try and it seems to work:

[build.log](https://github.com/ANTsX/ANTs/files/3772758/build.log)
[install.log](https://github.com/ANTsX/ANTs/files/3772763/install.log)
[mytree.txt](https://github.com/ANTsX/ANTs/files/3772765/mytree.txt)

gdevenyi:
Hi,

The build you logged did not use -DBUILD_SHARED_LIBS so the shared libraries were not installed and so that codepath was not triggered.

This line needs to be updated to conform with the lib vs lib64 detection inside CMAKE
https://github.com/ANTsX/ANTs/blob/master/ANTS.cmake#L178

In addition, the ITK/VTK installs may need to be adjusted so that the staging superbuild install honours lib/lib64, details on the Gentoo ITK/VTK configuration would be helpful here to see how thats handled.

cookpa:
Addressed in #1062

PR #886, opened 2019-10-17
COMP:  GCC compilation warnings round 2.

ntustison:


gdevenyi:
Timeouts need a restart, otherwise looks good.

PR #907, opened 2019-11-04
COMP:  Round 9

ntustison:


gdevenyi:
Awesome.

Purely from an infrastructure perspective, would it be possible for you to do your branch/dev on a personal account fork and then PR to here? Because you've created/pushed a branch here and also opened a PR, we actually get two travis builds, which slows things down if there are other PRs waiting around.

Just a thought :)

cookpa:
Can we configure travis to only test master + PRs to master?

gdevenyi:
Possibly. 

Looks like I can add it to the travis code, will test:
https://docs.travis-ci.com/user/customizing-the-build/#safelisting-or-blocklisting-branches

PR #915, opened 2019-11-19
COMP: Install ANTs with the make command

cookpa:
This installs to CMAKE_INSTALL_PREFIX after building, removes need for

cd ANTS-build
make install

### DIFF
--- a/SuperBuild.cmake
+++ b/SuperBuild.cmake
@@ -297,7 +297,6 @@ ExternalProject_Add(${proj}
     ${CMAKE_OSX_EXTERNAL_PROJECT_ARGS}
     ${COMMON_EXTERNAL_PROJECT_ARGS}
     -D${LOCAL_PROJECT_NAME}_SUPERBUILD:BOOL=OFF
-  INSTALL_DIR ${CMAKE_INSTALL_PREFIX}
   )
 
 ## Force rebuilding of the main subproject every time building from super structure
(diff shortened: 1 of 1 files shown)


cookpa:
Thoughts? This installs during the superbuild `make`, which is also not completely standard, but I think it would overall cut down on user confusion?

gdevenyi:
Well this is easy, I was hoping to do something about this.

What do the logs look like when you make install?

gdevenyi:
And, have you exercised all combination of BUILD_SHARED_LIBS and use system VTK/ITK settings?

cookpa:
It looks like I was doing it wrong; one should not set `INSTALL_DIR` because that doesn't actually configure anything, it just changes the `<INSTALL_DIR>` token that can be expanded by the external project if it wants.

The important change was to stop setting `INSTALL_COMMAND=""`, because that disables the install step.

https://cmake.org/cmake/help/latest/module/ExternalProject.html

The problem with this approach is that if we want to install to a privileged location, we have to run the whole make with those privileges, and not just `make install`.

cookpa:
Going to close for now, as I need to work on other things. It will take some time to test various combinations of system ITK, shared libs, etc.

gdevenyi:
I have to wrap my head around how the minc-toolkit-v2 superbuild does this, as make and make install work at the top level fine.

PR #916, opened 2019-11-19
BUG: Restoring semi-colon at end of warning macro

hjmjohnson:
The semi-colon was recently removed to address a
compiler warning about duplicate `;`.  This warning
was also addressed in upstream ITK, and now the the
`;` is required to compile with ITK versions after 11/06/2019.


ntustison:
Thanks @hjmjohnson .  Are these the only instances that need to be reversed?   I thought I had removed more.

hjmjohnson:
@ntustison This is all that I found.  Build succeeds.

gdevenyi:
Thanks for the catch @hjmjohnson 

I think I'll try to hack up a travis job that builds master against the ITK HEAD nightly or somesuch, so we can get out ahead of these.

hjmjohnson:
Failures are related to timeouts of build system, not code problems.

ntustison:
Thanks @hjmjohnson

PR #921, opened 2019-12-04
lightweight tags are found in git describe

muschellij2:
This should now look for lightweight tags and also annotated tags, not just annotated tags.  This was causing the issue 

https://github.com/ANTsX/ANTsRCore/issues/15

This is a dirty hack to fix that issue, as the git repo is pointing to ANTsRCore and not ANTs itself, which it should be.  
Related:
https://github.com/rpavlik/cmake-modules/issues/35

Therefore, one tag in the last 50 (as Travis does a git clone with `--depth=50`) has to have the form of `vN.N.N` or something of the like for this to work (always starting with `v`).  Currently, this only works when the tag is an annotated tag.  This fix will work for all tags.


cookpa:
Restarting the failed Travis builds

muschellij2:
ANy thoughts on this?

gdevenyi:
I'm not quite sure what this addresses in terms of how the error arises yet. Do we have a complete view?

Re: Travis failing, I've made some improvements since, can you rebase on master so we can build again?

muschellij2:
The issue occurs is others trying to build but not on the ANTsX username, see https://github.com/ANTsX/ANTsRCore/issues/15 for an example.  The way the checks are done for the tag is that it looks for an annotated tag (or any tag with my fix), which is constrained to the last 50 commits as --depth=50 is done on Travis and other systems.   Using any tag (not annotated) fixes this.

gdevenyi:
Ah, I didn't realize there was another solution to the shallow clone. I fixed that a while ago here via https://github.com/ANTsX/ANTs/blob/master/.travis.yml#L10-L11.

If this produces the same results in all cases but avoids the bug, I'm fine with it.

PR #926, opened 2020-01-06
antsApplyTransforms: add option -odt to set output datatype

CGSchwarzMayo:
Adds an antsApplyTransforms option --odt to set output datatype, via itkCastImageFilter before writing the output image to disk (internal computations remain in double or float regardless). Our typical use case applies antsApplyTransforms to propagate a library of atlases where values are integers <255, so there are large gains in disk space efficiency and I/O time by writing these as uchar rather than double. It would be possible to convert externally after writing, but this would not be I/O efficient.

     --odt char
           uchar
           short
           int
           float
           double
           default
          Output image data type. This is a direct typecast; output values are not 
          rescaled. Default is to use the internal data type (float or double). uchar is 
          unsigned char; others are signed. WARNING: Outputs will be incorrect if values 
          exceed the range allowed by your choice. Note that some pixel types are not 
          supported by some image formats. e.g. int is not supported by jpg.

### DIFF
--- a/Examples/antsApplyTransforms.cxx
+++ b/Examples/antsApplyTransforms.cxx
@@ -894,7 +894,6 @@ static void antsApplyTransformsInitializeCommandLineOptions( itk::ants::CommandL
   OptionType::Pointer option = OptionType::New();
   option->SetLongName( "output-data-type" );
   option->SetShortName( 'u' );
-  option->SetUsageOption( 0, "Linear" );
   option->SetUsageOption( 0, "char" );
   option->SetUsageOption( 1, "uchar" );
   option->SetUsageOption( 2, "short" );
--- a/Examples/antsApplyTransforms.cxx
+++ b/Examples/antsApplyTransforms.cxx
@@ -886,8 +886,8 @@ static void antsApplyTransformsInitializeCommandLineOptions( itk::ants::CommandL
     + std::string( "This is a direct typecast; output values are not rescaled. " )
     + std::string( "Default is to use the internal data type (float or double). " )
     + std::string( "uchar is unsigned char; others are signed. " )
-    + std::string( "WARNING: Outputs will be incorrect if values exceed the " )
-    + std::string( "range allowed by your choice. " )
+    + std::string( "WARNING: Outputs will be incorrect (overflowed/reinterpreted) " )
+    + std::string( "if values exceed the range allowed by your choice. " )
     + std::string( "Note that some pixel types are not supported by some image " )
     + std::string( "formats. e.g. int is not supported by jpg. " );
 
--- a/Examples/antsApplyTransforms.cxx
+++ b/Examples/antsApplyTransforms.cxx
@@ -9,6 +9,7 @@
 #include "itkResampleImageFilter.h"
 #include "itkVectorImage.h"
 #include "itkVectorIndexSelectionCastImageFilter.h"
+#include "itkCastImageFilter.h"
 
 #include "itkAffineTransform.h"
 #include "itkCompositeTransform.h"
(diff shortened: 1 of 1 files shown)


ntustison:
@CGSchwarzMayo Thanks for this.  I'll look a little closer once the travis builds finish but, at first glance, please write out the long variable name, i.e., "output-data-type" instead of "odt".

gdevenyi:
Yes please! I've been missing this since moving my pipelines from minc land.

Examples/antsApplyTransforms.cxx
```diff
@@ -849,6 +880,30 @@ static void antsApplyTransformsInitializeCommandLineOptions( itk::ants::CommandL
   parser->AddOption( option );
   }
 
+  {
+  std::string description =
+    std::string( "Output image data type. " )
+    + std::string( "This is a direct typecast; output values are not rescaled. " )
+    + std::string( "Default is to use the internal data type (float or double). " )
+    + std::string( "uchar is unsigned char; others are signed. " )
+    + std::string( "WARNING: Outputs will be incorrect if values exceed the " )
```
gdevenyi: I think perhaps this should be more explicit, how is it incorrect? Clipped or rescaled?

CGSchwarzMayo:
> @CGSchwarzMayo Thanks for this. I'll look a little closer once the travis builds finish but, at first glance, please write out the long variable name, i.e., "output-data-type" instead of "odt".

Okay, done! Thanks for your consideration.

Examples/antsApplyTransforms.cxx
CGSchwarzMayo: > I think perhaps this should be more explicit, how is it incorrect? Clipped or rescaled?

I changed the text to ..."incorrect (overflowed/reinterpreted)"... 
How's that?

It's not clipped or rescaled. It just goes through a raw typecast. I believe the behavior of particular some typecasts is defined and others are undefined or implementation-dependent, so there is not a short phrase to fully describe this. "overflowed/reinterpreted" was the best I could come up with.

Examples/antsApplyTransforms.cxx
```diff
@@ -849,6 +880,32 @@ static void antsApplyTransformsInitializeCommandLineOptions( itk::ants::CommandL
...
+    + std::string( "uchar is unsigned char; others are signed. " )
+    + std::string( "WARNING: Outputs will be incorrect (overflowed/reinterpreted) " )
+    + std::string( "if values exceed the range allowed by your choice. " )
+    + std::string( "Note that some pixel types are not supported by some image " )
+    + std::string( "formats. e.g. int is not supported by jpg. " );
+
+  OptionType::Pointer option = OptionType::New();
+  option->SetLongName( "output-data-type" );
+  option->SetShortName( 'u' );
+  option->SetUsageOption( 0, "Linear" );
```
gdevenyi: Somehow this snuck in?

CGSchwarzMayo: The short name? Yes, I mentioned that in the commit message. --output-data-type is very long. It seemed reasonable to choose a short name also, since everything else except for --float has one.

gdevenyi: `` option->SetUsageOption( 0, "Linear" );``

CGSchwarzMayo: Whoops. Yes, that was just a paste error when adding the short name. Removed, thanks.

ntustison:
@CGSchwarzMayo Thanks.  This looks good to me.  I'll defer to @gdevenyi concerning any additional issues.  But, otherwise, as soon as the travis builds finish green, we can go ahead and merge.

tashrifbillah:
(y)

Issue #931, opened 2020-01-10
antsMotionCorr -- add support for other interpolation strategies


dangom:
**Is your feature request related to a problem? 

I noticed the resampled images that come as output of antsMotionCorr are somewhat smooth. Checking the code it seems like the ResampleImageFilter interpolator being used is the default   LinearInterpolateImageFunction, and users cannot change this.

**Describe the solution you'd like**
antsMotionCorr expose other interpolation schemes via a `--interpolation` flag.

**Describe alternatives you've considered**
I've considered saving the transform as a warpfield, and then applying it to the BOLD series later using antsApplyTransforms choosing a different interp algorithm. That is cumbersome in that it requires first replicating the reference to make it 4D, and there is always room for error when calling two commands instead of a single one.

dangom:
I guess somewhere here:
https://github.com/ANTsX/ANTs/blob/f19c6394c35108c373ca8643814f9a37ea45e752/Examples/antsMotionCorr.cxx#L1428

A line with:
```
resampler->SetInterpolator( interpolator );
```

where `interpolator` comes from a command line flag?

ntustison:
@dangom --- this sounds like a good idea.  Do you feel comfortable putting together a pull request?

dangom:
I would love to. So far I've figured out that I have to add the following:

0. Add relevant headers:

```
#include "itkBSplineInterpolateImageFunction.h"
#include "itkLinearInterpolateImageFunction.h"
#include "itkGaussianInterpolateImageFunction.h"
#include "itkInterpolateImageFunction.h"
#include "itkNearestNeighborInterpolateImageFunction.h"
#include "itkWindowedSincInterpolateImageFunction.h"
#include "itkLabelImageGaussianInterpolateImageFunction.h"
#include "itkLabelImageGenericInterpolateImageFunction.h"

```

1. Add the interpolation options to the cli parser:
```
  {
    std::string description =
      std::string( "Several interpolation options are available in ITK. " )
      + std::string( "These have all been made available." );

    OptionType::Pointer option = OptionType::New();
    option->SetLongName( "interpolation" );
    option->SetShortName( 'n' );
    option->SetUsageOption( 0, "Linear" );
    option->SetUsageOption( 1, "NearestNeighbor" );
    option->SetUsageOption( 2, "Gaussian[<sigma=imageSpacing>,<alpha=1.0>]" );
    option->SetUsageOption( 3, "BSpline[<order=3>]" );
[...]

PR #932, opened 2020-01-11
Add support for an --interpolation flag in antsMotionCorr

dangom:
Note that no support was added for either Gaussian or genericLabel
interpolation routines, as those require adapting parameters according
to the voxel dimensions. If users want these, they can always fall
back to using antsApplyTransforms for a more generic solution.

### DIFF
--- a/Examples/antsMotionCorr.cxx
+++ b/Examples/antsMotionCorr.cxx
@@ -69,6 +69,13 @@
 #include "itkSimilarity2DTransform.h"
 #include "itkSimilarity3DTransform.h"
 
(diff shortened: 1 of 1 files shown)


dangom:
Closes #931

Ready for review

ntustison:
Thanks @dangom .  We'll let the travis builds complete and then we'll merge.

dangom:
No prob.
Also worth mentioning that I had to change the short name of --interpolation to -p because -n is already in use by --n-images.

ntustison:
@dangom Thanks again.

Issue #940, opened 2020-01-22
ANTs + system ITK 5.1


crossmanith:
In order to build with system ITK 5.1 I had to change FileModeType to FileModeEnum and TransformCategoryType to TransformCategoryEnum. I've attached two patches instead of build logs because finally I could build ANTs successfully.

**When did the error occur?**
[ ] CMake configuration (cmake / ccmake)
[x] Compilation (make)
[ ] Installation (make install)

**Build environment**
 - OS: Linux (Ubuntu)
 - OS version: 4.4.0-170-generic
 - Type of system: [Desktop, laptop, HPC cluster, cloud instance,
   other] desktop
 
**ANTs version**
commit a025d04

[0001-TransformType-TransformCategoryType-TransformType-Tr.patch.txt](https://github.com/ANTsX/ANTs/files/4098117/0001-TransformType-TransformCategoryType-TransformType-Tr.patch.txt)
[0002-FileModeType-FileModeEnum.patch.txt](https://github.com/ANTsX/ANTs/files/4098118/0002-FileModeType-FileModeEnum.patch.txt)

ntustison:
Can you convert this to a pull request?

crossmanith:
Some instructions how I can achieve this quickly?

crossmanith:
fork + push to my fork and then new pull request?

ntustison:
Yes

crossmanith:
Done

ntustison:
Thanks.  We'll wait for the travis builds before we merge.

crossmanith:
As far as I can see, the first config in travis builds successfully but in the end there is some "deadline error message" which causes a failure for the build process. Anything I can do?

PR #941, opened 2020-01-22
Use Enum instead of Type

crossmanith:


gdevenyi:
Builds all fail on ANTs.

Does your update for ITK5.1 require a newer version ITK?

Re: #940, these are not timeouts, these are failures.

gdevenyi:
It would be helpful if you could provide more version information for your "system ITK 5.1"

crossmanith:
My ITK tree has 8d3961 as the most recent commit. I'll update and see if my patch still works.

ntustison:
@crossmanith -- Did you update the ITK git tag [here](https://github.com/ANTsX/ANTs/blob/master/SuperBuild/External_ITKv5.cmake#L154)?

crossmanith:
Commit 21fc5c94b58c00f5db1a8ff4d2a18e2fbeb59014 from 12/29/2019 introduced the Enums, ANTs checks out an earlier ITK version from the end of november. This is why the travis builds are failing.

crossmanith:
I've modified the ITK git tag, set USE_SYSTEM_ITK to OFF and see if it builds successfully

ntustison:
@crossmanith Thanks for checking.

crossmanith:
Do I have to find the earliest ITK commit for a successful build of ANTs or can I take any working commit?

ntustison:
On something like this, we would simply use the most recent commit.

gdevenyi:
Before we merge this, should perhaps we hold on this until there's a need to bump the version again? 

I don't immediately seem any changes in https://github.com/InsightSoftwareConsortium/ITK/compare/49671655c88612fabeb5f0784d3f40123100f166...master that would impact ANTs, except for perhaps https://github.com/InsightSoftwareConsortium/ITK/pull/1557

ntustison:
I would say that this is sufficient reason to bump the ITK version.  We always try to stay current with the latest ITK changes.

gdevenyi:
@crossmanith You can just add the tag change as a new commit on this branch and then git push to update this PR.

PR #943, opened 2020-01-29
adjust ITK git tag

crossmanith:
I've chose the ITK git tag which is most recent in my local installation - should build now with patch type->enum

Do I have to include code changes and cmake changes in the same pull request?


crossmanith:
trying with a single patch for source code changes and change in External_ITK.cmake

PR #946, opened 2020-02-04
Type->Enum // added .as_ref() // adjust ITK git tag

crossmanith:


crossmanith:
Looks like some builds fail due to a time out - how can we avoid that?

gdevenyi:
I'll take a look today

gdevenyi:
We're at the hairy edge of timeout, the build succeeds but we run out of time during the caching.

I've updated the travis config at #946 to ensure that the cache gets populated, which should solve this.

From my perspective, the changes look ok to merge.

ntustison:
Sounds good.  Thanks @crossmanith and @gdevenyi .

PR #947, opened 2020-02-05
Update travis

gdevenyi:


gdevenyi:
Attempt to address timeout in #946

cookpa:
Does this need restarting?

gdevenyi:
We're still bumping up again 1h 30min max. I need to cut another 5 min off the timeout wrapper so that cache updating succeeds. Will force-push a change to trigger a restart.

gdevenyi:
Looking good here to avoid timeouts and populating the cache.

cookpa:
Thanks

Issue #914, opened 2019-11-19
Docker container with ANTs 


dorianps:
I have created an ANTs docker container and a few scripts to help the process. I am not using much ANTs these days, so if anyone wants to give it a try to process anything, it will be good to know. I tried `PrintHeader` and works as expected.
https://github.com/dorianps/docker/tree/master/ants

The container is build with multi-staging to reduce the size. Total size now is 2.5Gb (from 6Gb if developer tools were left in the container)

*Sorry for not using the issue template, did not seem to fit the topic.*

gdevenyi:
Looks good @dorianps . You can rm -rf /opt/ANTs/lib since that's only installed because of how the ANTsR build assumes things right now. That should nicely shrink the build.

satra:
just an fyi that we have been building/distributing containers containing ants from source via [neurodocker](https://github.com/kaczmarj/neurodocker/)  

the current ants download/build spec is here: https://github.com/kaczmarj/neurodocker/blob/master/neurodocker/templates/ants.yaml

gdevenyi:
Hi @satra that script will be broken since the improvements of cmake to properly honour CMAKE_INSTALL_PREFIX, @dorianps 's version does it the supported way now.

satra:
@gdevenyi - is this a cmake issue or an ants issue? i.e. which software's version should we be checking?

ping: @kaczmarj

cookpa:
It's an ANTs issue, it's overall much better IMO. The only annoying thing is that you have to cd to `ANTS-build` to run `make install`. But the make install creates bin/ and lib/ under `$CMAKE_INSTALL_PREFIX`, and copies the scripts too (hooray)

gdevenyi:
ANTs' cmake scripts were updated as of https://github.com/ANTsX/ANTs/pull/841 to stop hard-coding the bin/ and lib/ directories, so that "make install" will work.

A bug in that config was fixed at https://github.com/ANTsX/ANTs/pull/867 to complete the conversion.

satra:
@cookpa and @gdevenyi  - that's great. we will check the git commit hash to determine if we have to use the older or newer form.

dorianps:
Thanks for the input.

@gdevenyi  Your tip shaves another 500mb from the container. Currently the container is 2Gb of which 1.8Gb is /opt/ANTs/bin and 0.2Mb is the system itself.
[...]

PR #948, opened 2020-02-05
Automatically build Docker Images

mattcieslak:
This PR adds an automatic Docker image build on CircleCI.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,56 @@
+# Use Ubuntu 16.04 LTS
+FROM nvidia/cuda:9.1-runtime-ubuntu16.04
+
+# Prepare environment
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+                    curl \
+                    bzip2 \
+                    ca-certificates \
+                    build-essential \
+                    autoconf \
+                    libtool \
+                    pkg-config \
+                    wget \
+                    libboost-all-dev \
+                    unzip \
+                    libgl1-mesa-dev \
+                    libglu1-mesa-dev \
+                    freeglut3-dev \
+                    mesa-utils \
+                    g++ \
+                    jq \
+                    tar \
+                    zip \
+                    gcc \
+                    make \
+                    python \
+                    zlib1g-dev \
+                    imagemagick \
+                    software-properties-common \
+                    git && \
+    curl -sL https://deb.nodesource.com/setup_10.x | bash - && \
+    apt-get install -y --no-install-recommends \
+      nodejs && \
+    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+ADD https://cmake.org/files/v3.11/cmake-3.11.4-Linux-x86_64.sh /cmake-3.11.4-Linux-x86_64.sh
+
+ADD . /tmp/ants/source
+RUN mkdir /opt/cmake \
+  && sh /cmake-3.11.4-Linux-x86_64.sh --prefix=/opt/cmake --skip-license \
+  && ln -s /opt/cmake/bin/cmake /usr/local/bin/cmake \
+    && cd /tmp/ants/source \
+    && mkdir -p /tmp/ants/build \
+    && cd /tmp/ants/build \
+    && mkdir -p /opt/ants-latest \
+    && git config --global url."https://".insteadOf git:// \
+    && cmake -DBUILD_TESTING=OFF -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX=/opt/ants-latest /tmp/ants/source \
+    && make -j2 \
+    && cd ANTS-build \
+    && make install \
+    && rm -rf /tmp/ants \
+    && rm -rf /opt/cmake /usr/local/bin/cmake
+
+ENV ANTSPATH="/opt/ants-latest/bin" \
+    PATH="/opt/ants-latest/bin:$PATH" \
+    LD_LIBRARY_PATH="/opt/ants-latest/lib:$LD_LIBRARY_PATH"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,54 +1,39 @@
-# Use Ubuntu 16.04 LTS
-FROM nvidia/cuda:9.1-runtime-ubuntu16.04
+FROM ubuntu:bionic-20200112 as builder
 
-# Prepare environment
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
-                    curl \
-                    bzip2 \
-                    ca-certificates \
-                    build-essential \
-                    autoconf \
-                    libtool \
-                    pkg-config \
-                    wget \
-                    libboost-all-dev \
-                    unzip \
-                    libgl1-mesa-dev \
-                    libglu1-mesa-dev \
-                    freeglut3-dev \
-                    mesa-utils \
-                    g++ \
-                    jq \
-                    tar \
-                    zip \
-                    gcc \
-                    make \
-                    python \
-                    zlib1g-dev \
-                    imagemagick \
                     software-properties-common \
-                    git && \
-    curl -sL https://deb.nodesource.com/setup_10.x | bash - && \
-    apt-get install -y --no-install-recommends \
-      nodejs && \
-    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
-ADD https://cmake.org/files/v3.11/cmake-3.11.4-Linux-x86_64.sh /cmake-3.11.4-Linux-x86_64.sh
+                    build-essential \
+                    apt-transport-https \
+                    ca-certificates \
+                    gnupg \
+                    software-properties-common \
+                    wget \
+                    ninja-build \
+                    git \
+                    zlib1g-dev
+
+RUN wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null \
+    | apt-key add - \
+    && apt-get update \
+    && apt-get -y install cmake
 
 ADD . /tmp/ants/source
-RUN mkdir /opt/cmake \
-  && sh /cmake-3.11.4-Linux-x86_64.sh --prefix=/opt/cmake --skip-license \
-  && ln -s /opt/cmake/bin/cmake /usr/local/bin/cmake \
-    && mkdir -p /tmp/ants/build \
+RUN mkdir -p /tmp/ants/build \
     && cd /tmp/ants/build \
     && mkdir -p /opt/ants-latest \
     && git config --global url."https://".insteadOf git:// \
-    && cmake -DBUILD_TESTING=OFF -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX=/opt/ants-latest /tmp/ants/source \
+    && cmake \
+      -DBUILD_TESTING=OFF \
+      -DBUILD_SHARED_LIBS=ON \
+      -DCMAKE_INSTALL_PREFIX=/opt/ants-latest \
+      /tmp/ants/source \
     && make -j2 \
     && cd ANTS-build \
-    && make install \
-    && rm -rf /tmp/ants \
-    && rm -rf /opt/cmake /usr/local/bin/cmake
+    && make install
+
+FROM ubuntu:bionic-20200112
+COPY --from=builder /opt/ants-latest /opt/ants-latest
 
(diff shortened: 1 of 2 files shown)


cookpa:
Thanks @mattcieslak!

@ntustison @stnava @gdevenyi Matt is going to help me set up the free circleCI / Dockerhub routines to support this

ntustison:
Thanks!

cookpa:
Looks like Travis is going to take a while, so we got started with the circleci / Docker config. I made an antsx org on DockerHub, let me know your handles over there and I can add you

https://hub.docker.com/orgs/antsx

gdevenyi:
Hi,

This looks cool, but without cleanup we'll have a huge docker file with all the build dependencies in it for no reason. There's some other dockers for ANTs already out there, this guy is doing the right thing with build-then-copy to avoid having a "thick" container:
https://github.com/kaczmarj/ANTs-builds

Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+
+ADD . /tmp/ants/source
+RUN mkdir /opt/cmake \
+  && sh /cmake-3.11.4-Linux-x86_64.sh --prefix=/opt/cmake --skip-license \
+  && ln -s /opt/cmake/bin/cmake /usr/local/bin/cmake \
+    && mkdir -p /tmp/ants/build \
+    && cd /tmp/ants/build \
+    && mkdir -p /opt/ants-latest \
+    && git config --global url."https://".insteadOf git:// \
+    && cmake -DBUILD_TESTING=OFF -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX=/opt/ants-latest /tmp/ants/source \
```
gdevenyi: I've always been a bit confused for shared vs static. The default ants build is static, which means all that's needed is the binaries at the end of everything.

Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+                    tar \
+                    zip \
+                    gcc \
+                    make \
+                    python \
+                    zlib1g-dev \
+                    imagemagick \
+                    software-properties-common \
+                    git && \
+    curl -sL https://deb.nodesource.com/setup_10.x | bash - && \
```
gdevenyi: pretty sure we don't need node

Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+    apt-get install -y --no-install-recommends \
+      nodejs && \
+    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+ADD https://cmake.org/files/v3.11/cmake-3.11.4-Linux-x86_64.sh /cmake-3.11.4-Linux-x86_64.sh
```
gdevenyi: See my travis config for an offical cmake PPA source from kitware

Dockerfile
cookpa: Shared libs make for smaller containers. They are less convenient as you have to worry about having the libraries loaded, but I think in a container this can be solved once. That was my take from https://github.com/ANTsX/ANTs/issues/914

Dockerfile
cookpa: @mattcieslak do you think it's worth trimming the package list a bit?

mattcieslak: many of these can go. They came with on a copy/paste from another dockerfile

Dockerfile
mattcieslak: > This looks cool, but without cleanup we'll have a huge docker file with all the build dependencies in it for no reason. There's some other dockers for ANTs already out there, this guy is doing the right thing with build-then-copy to avoid having a "thick" container:
> https://github.com/kaczmarj/ANTs-builds

I just split up the Dockerfile into two parts and that definitely makes a smaller image. I wasn't aware of that trick.

Dockerfile
```diff
@@ -1,54 +1,39 @@
...
-    apt-get install -y --no-install-recommends \
-      nodejs && \
-    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
-ADD https://cmake.org/files/v3.11/cmake-3.11.4-Linux-x86_64.sh /cmake-3.11.4-Linux-x86_64.sh
+                    wget \
+                    ninja-build \
+                    git \
+                    zlib1g-dev
+
+RUN wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null \
```
gdevenyi: Needs apt-add-repository -y 'deb https://apt.kitware.com/ubuntu/ bionic main'

Dockerfile
```diff
@@ -1,54 +1,39 @@
...
-    && mkdir -p /tmp/ants/build \
+RUN mkdir -p /tmp/ants/build \
     && cd /tmp/ants/build \
     && mkdir -p /opt/ants-latest \
     && git config --global url."https://".insteadOf git:// \
-    && cmake -DBUILD_TESTING=OFF -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX=/opt/ants-latest /tmp/ants/source \
+    && cmake \
+      -DBUILD_TESTING=OFF \
+      -DBUILD_SHARED_LIBS=ON \
+      -DCMAKE_INSTALL_PREFIX=/opt/ants-latest \
```
gdevenyi: I think I'd prefer /opt/ants/VERSION as the installation directory.

Dockerfile
```diff
@@ -1,54 +1,39 @@
...
-RUN mkdir /opt/cmake \
-  && sh /cmake-3.11.4-Linux-x86_64.sh --prefix=/opt/cmake --skip-license \
-  && ln -s /opt/cmake/bin/cmake /usr/local/bin/cmake \
-    && mkdir -p /tmp/ants/build \
+RUN mkdir -p /tmp/ants/build \
     && cd /tmp/ants/build \
     && mkdir -p /opt/ants-latest \
     && git config --global url."https://".insteadOf git:// \
-    && cmake -DBUILD_TESTING=OFF -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX=/opt/ants-latest /tmp/ants/source \
+    && cmake \
```
gdevenyi: Needs "-GNinja"

Dockerfile
```diff
@@ -1,54 +1,39 @@
...
+      -DBUILD_TESTING=OFF \
+      -DBUILD_SHARED_LIBS=ON \
+      -DCMAKE_INSTALL_PREFIX=/opt/ants-latest \
+      /tmp/ants/source \
     && make -j2 \
```
gdevenyi: Needs "cmake --build . --parallel" to use Ninja

Dockerfile
```diff
@@ -1,54 +1,39 @@
...
     && cd ANTS-build \
-    && make install \
-    && rm -rf /tmp/ants \
-    && rm -rf /opt/cmake /usr/local/bin/cmake
+    && make install
```
gdevenyi: Replace with "cmake --install ."

Dockerfile
```diff
@@ -1,54 +1,39 @@
...
+
+FROM ubuntu:bionic-20200112
+COPY --from=builder /opt/ants-latest /opt/ants-latest
 
 ENV ANTSPATH="/opt/ants-latest/bin" \
     PATH="/opt/ants-latest/bin:$PATH" \
```
gdevenyi: The use of -DBUILD_SHARED_LIBS=ON  means you need to define LD_LIBRARY_PATH for the libs dir.

Dockerfile
cookpa: Or simply `/opt/ants`? The user has already decided the version to use when they download a particular container image

gdevenyi:
Okay, looking good so far, my next concern is that we will timeout with circleCI since ITK/ANTs takes a long time to build.

I presume circleCI has a caching facility of somesort? Can we try to bind-mound a circleCI ccache into the container and enable ccache in the container?

Dockerfile
mattcieslak: Is there a place other than in the last ENV line where this should be specified?

mattcieslak:
> Okay, looking good so far, my next concern is that we will timeout with circleCI since ITK/ANTs takes a long time to build.
> 
> I presume circleCI has a caching facility of somesort? Can we try to bind-mound a circleCI ccache into the container and enable ccache in the container?

CircleCI won't time out. We regularly do 3+ hour builds on there. You have to pay for the Docker layer caching, so the only alternative is to `docker export | tar` which takes more time than just starting from scratch.

gdevenyi:
> CircleCI won't time out. We regularly do 3+ hour builds on there.

Oh wow that's great. 1.5hr on travis has been killing me. I will have to investigate if we can transition out testing infrastructure there...

> You have to pay for the Docker layer caching, so the only alternative is to docker export | tar which takes more time than just starting from scratch.

I'm not referring to docker caching, but rather manual directory caching as per: https://circleci.com/docs/2.0/caching/

If we define a directory for ccache to store files, bind it inside the docker container,  and we save/restore it on circleci shutdown/startup, we can cache a huge amount of precompiled objects. This works very well on the travis caching saving almost an hour.

Dockerfile
gdevenyi: Thats the right place, sorry if I missed that.

dorianps:
Cool, what is the container size that is coming out?

Since you are at it, can you extend this to build the singularity container from docker? It is likely that any user in shared environments will go straight for singularity as docker is not an option in shared environments due to security risks.

Another pet-peeve, check if you can print a few lines on the console to inform the user about the version, like I do here:
https://github.com/dorianps/docker/tree/master/ants#windows-runantswindowsbat-double-click

mattcieslak:
> Cool, what is the container size that is coming out?
> 
> Since you are at it, can you extend this to build the singularity container from docker? It is likely that any user in shared environments will go straight for singularity as docker is not an option in shared environments due to security risks.
> 
> Another pet-peeve, check if you can print a few lines on the console to inform the user about the version, like I do here:
> https://github.com/dorianps/docker/tree/master/ants#windows-runantswindowsbat-double-click

The image comes out at 382MB. I'm not familiar with distributing singularity images, we usually `singularity build` from images on Dockerhub. Maybe in another PR for these suggestions once this gets merged?

dorianps:
I have not built or pushed containers using automated online services. I usually use docker2singularity locally to convert a docker image to singularity, which works rather well. 
If there is a way to run that single command somewhere online that may be good (my docker hub pushes are done on my local server).  Singularity is also supposed to have a hub, but I have not used it before.

cookpa:
Once the docker is complete I can take a crack at having it build over on SingularityHub. The trick will be triggering builds from DockerHub as opposed to Github

mattcieslak:
I believe we can build a singularity image directly from the local docker image created on circleci and push it to singularity hub right after it gets pushed to dockerhub. This should probably go in another PR though

mattcieslak:
Looks like the builds are timing out for XCode.

cookpa:
Yeah I've restarted them a bunch of times, I think they are not going to work. 

Are we otherwise ready to merge?

ntustison:
I'm fine with merging.

gdevenyi:
Before this merges, is it possible to run some kind of quick test on the docker image by invoking one of the programs against some test data?

Re: timeouts, we can ignore that, since this is orthogonal to the travis builds.

cookpa:
Adding a test is a good idea. Merging so we can start using the container.

gdevenyi:
I don't see the repo at https://hub.docker.com/u/antsx, is the container downloadable yet?

cookpa:
I need to set up some keys...might take me a few days

Sent from my iPhone

On Feb 13, 2020, at 10:24 AM, Gabriel A. Devenyi <notifications@github.com> wrote:

﻿

I don't see the repo at https://hub.docker.com/u/antsx, is the container downloadable yet?

cookpa:
OK I think I can't do this because I'm not an admin of the ANTsX organization here on Github. I was able to create the ANTsX org on circleci, but an admin from here will need to join there to do it

https://circleci.com/docs/2.0/gh-bb-integration/

ntustison:
Okay, I think it's approved.

cookpa:
Thanks Nick! I see it is approved, but I'm still having problems. Would you mind trying to enable the "deploy key" on circleci? 

https://circleci.com/gh/ANTsX/ANTs/444

Where it says 

"Project ANTsX/ANTs isn't configured with a deploy key or a github user, so we may not be able to test all pushes."

There's a button to add a key. 

Alternatively I think you can do it here

https://circleci.com/gh/ANTsX/ANTs/edit#checkout

ntustison:
Okay, see if that does it.

PR #963, opened 2020-03-06
Sharpening options in ImageMath, updates to AverageImages, template scripts

cookpa:
New features:

`ImageMath` UnsharpMask command, which provides access to this filter

https://itk.org/Doxygen/html/classitk_1_1UnsharpMaskImageFilter.html

and its main parameters: amount (default 0.5, useful range 0-2), radius (default 1mm) and threshold (default 0).

`AverageImages` now has three options:

0 - just get the mean in each voxel
1 - mean of each voxel, but input volumes are divided by the mean within the volume, then Laplacian sharpening is applied to the result
2 - mean of each voxel, but input volumes are divided by the mean within the volume. No sharpening

This allows users to sharpen using their choice of method with ImageMath, or have backwards compatibility by choosing option 0 or 1.

In `antsMultivariateTemplateConstruction.sh` (and 2.sh), I've left Laplacian sharpening for the initial / rigid template steps, as a more aggressive approach is probably a good thing there. For the main template update, I switched to UnsharpMask with default parameters. 

Depending on the application one might do better with a larger radius or larger amount, but this will likely be data dependent. With amount = 0.75 and radius = 3mm, I was getting pretty good results but there was a small amount of halo artifact again, but less than with Laplacian. 

FYI the UnsharpMask filter can introduce negative intensities.

### DIFF
--- a/Examples/AverageImages.cxx
+++ b/Examples/AverageImages.cxx
@@ -44,24 +44,15 @@ int AverageImages1(unsigned int argc, char *argv[])
   typedef itk::ImageFileReader<ImageType>              ImageFileReader;
   typedef itk::ImageFileWriter<ImageType>              writertype;
 
+    
+  const int normalizei = std::stoi(argv[3]);
+    
+  if( ( normalizei < 0 ) || ( normalizei > 2 ) )
     {
-    const std::string temp(argv[1]);
-    if( !( ( temp == "2" ) || ( temp == "3" ) || ( temp == "4" ) ) )
-      {
-      std::cerr << "ERROR:  Dimension option must be 2 or 3 or 4, " << temp << "given" << std::endl;
-      return EXIT_FAILURE;
-      }
+    std::cerr << "ERROR:  Normalize option must be 0 or 1 or 2, " << normalizei << " given" << std::endl;
+    return EXIT_FAILURE;
     }
-    {
-    const std::string temp(argv[3]);
-    if( !( ( temp == "0" ) || ( temp == "1" )  ) )
-      {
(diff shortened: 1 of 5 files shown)


cookpa:
Still testing this but wanted to give Travis a head start.

gdevenyi:
> , radius (default 1mm) and threshold (default 0).
Is it actually mm or voxels? I'd much prefer a voxel-based radius so that the pipeline runs independently of the real-world resolution of the input files.

gdevenyi:
From the ITK code
> Set/Get Sigma values measured in the units of image spacing. Default: 1.0.

cookpa:
I think "units of image spacing" means mm.

gdevenyi:
I believe this current implementation would act very differently if I were say to supply mouse brains vs human brains (vs human brains at a different resolution). If radius were in voxels this would fix that.

cookpa:
Agreed. Maybe we can change the image math function to take a specification like 1x1x1vox or similar.

gdevenyi:
Agreed, I would prefer it work like SmoothImage's implementation, voxels by default with option for mm

cookpa:
OK now it defaults to voxels with an option to change that. Like SmoothImage

gdevenyi:
Changes look good, build failures are just timeouts so this is okay to merge.

cookpa:
Thanks, I think the code will merge without breaking anything, I just don't have much evidence that it will actually build better templates.

cookpa:
Going to close this PR until it can be tested further, but branch is there if anyone gets curious.

I will make a new one just for the new features to ImageMath and AverageImages

PR #970, opened 2020-03-18
ENH: Use as_matrix for explicit type matching.

hjmjohnson:


hjmjohnson:
@ntustison Could not build in debug mode due to missing ';'

ntustison:
Thanks @hjmjohnson

Issue #968, opened 2020-03-17
Morphological operations


ntustison:
Hey, just wondering if anybody knows anything about this particular implementation quirk of certain morphological operators in ``ImageMath``.  If one takes a look at the [relevant code](https://github.com/ANTsX/ANTs/blob/master/Examples/antsUtilities.h#L156-L190) in ``antsUtilities.h``, one can see that the dilate and erode operations are as expected but the open and close operations are composites of dilate and erode which actually produce different results from the open and close ITK filters.  For example, the main image below is the ``ImageMath`` version of "closing" (white = 1, black = 0) of the binary segmentation mask shown in red.

<img width="1080" alt="Screen Shot 2020-03-16 at 8 21 03 PM" src="https://user-images.githubusercontent.com/324811/76819066-79076480-67c4-11ea-8778-731c0d219ac7.png">

cookpa:
Can you upload your mask? I can't seem to reproduce this on 

```
antsRegistration --version
ANTs Version: 2.3.1.dev23-gc3466
Compiled: Sep 29 2019 13:40:29
```

ntustison:
Sure.  I'm guessing a pipeline issue but, regardless, I'm compiling a pull request now.  

```
$ antsRegistration --version
ANTs Version: 3.0.0.0.dev109-g8d278
Compiled: Sep  8 2019 09:22:03
```

[r16mask.nii.gz](https://github.com/ANTsX/ANTs/files/4346545/r16mask.nii.gz)

cookpa:
Oh  yes, now I see the problem. The crazy floating point values make it look normal when loaded as a segmentation in SNAP

ntustison:
@cookpa Thanks for checking this.

gdevenyi:
I have been using iMath for these sorts of operations. Which is considered the "supported" implementation?

ntustison:
If I remember correctly @jeffduda started ``iMath`` as a long term project to eventually subsume all of ``ImageMath`` functionality in a more organized fashion.  Obviously, that is going to take quite some time so, for the time being, both implementations are "supported."

PR #971, opened 2020-03-18
BUG:  Address incorrect output in morphological close/open.

ntustison:
https://github.com/ANTsX/ANTs/issues/968

### DIFF
--- a/Examples/antsUtilities.h
+++ b/Examples/antsUtilities.h
@@ -24,8 +24,13 @@
 #include "itkBinaryBallStructuringElement.h"
 #include "itkBinaryDilateImageFilter.h"
 #include "itkBinaryErodeImageFilter.h"
+#include "itkBinaryMorphologicalClosingImageFilter.h"
+#include "itkBinaryMorphologicalOpeningImageFilter.h"
+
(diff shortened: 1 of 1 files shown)


cookpa:
Does this restore the expected behavior that open and close are identical to the combinations of dilate and erode? 

It looks like it would, as the structuring element is the same.

ntustison:
@cookpa This is expected behavior and probably preferred since it actually uses the open and close ITK filters as opposed to the dilate/erode combination equivalents.  I didn't look at the internals but I wouldn't be surprised if the dedicated filters have implementation optimizations.

cookpa:
I agree it makes sense to use the dedicated filter. I was just thinking it would be good to ensure all the parameter choices are in place to ensure

ImageMath 3 dilate.nii.gz MD input.nii.gz $radius
ImageMath 3 closed.nii.gz ME dilate.nii.gz $radius

gives the same results as

ImageMath 3 closed.nii.gz MC input.nii.gz $radius

It looks like that is the case, but I haven't built the branch to check.

ntustison:
Agreed.  Although if the results weren't the same then I would imagine that would warrant an ITK bug report.

PR #974, opened 2020-03-19
COMP: Assigning non-static data member within const member

hjmjohnson:
ANTs/Examples/TimeSCCAN.cxx:866:32: warning:
object backing the pointer will be destroyed at the end of the full-expression [-Wdangling-gsl]
        const char *longName = ( ( *it )->GetLongName() ).c_str();
                               ^~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.

error: cannot assign to non-static data member within const member function 'operator()'


ntustison:
Thanks @hjmjohnson .  This is a minor change and all the builds are passing so far so I'm going to go ahead and merge.

PR #977, opened 2020-03-22
ENH: CMake option ANTS_INSTALL_LIBS_ONLY to not install binaries

cookpa:
@muschellij2 Does this accomplish what you wanted to do with https://github.com/ANTsX/ANTs/pull/782?


gdevenyi:
In my original cmake cleanup I had configured static builds to only install the binaries and not the libs. Since you're working on this, could you please re-add a config option so that only binaries are installed for such a configuration mode?

muschellij2:
Yes I think that it does - thanks for looking into this!

cookpa:
@gdevenyi I was trying to do this some time ago, but my attempts to turn off the libraries also resulted in the executables not getting installed.

Do you have a link to an example that builds bin and not libs?

gdevenyi:
My implementation was here, (which a later PR modified)
https://github.com/ANTsX/ANTs/pull/867/files#diff-33e15993d1f64c43353034633e3d6885R175-R176

This implementation will build the apps but not install them. It think it might be worthwhile to try to adjust the build system to not build the apps if they're not going to be installed anyways. I realize right now thats not at all how the code is structured.

cookpa:
> My implementation was here, (which a later PR modified)
> https://github.com/ANTsX/ANTs/pull/867/files#diff-33e15993d1f64c43353034633e3d6885R175-R176
> 
> This implementation will build the apps but not install them. It think it might be worthwhile to try to adjust the build system to not build the apps if they're not going to be installed anyways. I realize right now thats not at all how the code is structured.

I tried to see what to change with (from master):

`git diff b2c828ff8e5ce537d7096af04b086e965fede9e8 ANTS.cmake SuperBuild.cmake`

But it looks like your changes in the above commit are still there?

cookpa:
Just to be clear, what I'm doing is compiling with `make`, then `cd ANTS-build` and `make install`. I've got that last step producing libraries only with the options set.

But I could not figure out how to get `make install` to put executables into `bin/` but not libraries into `lib/`.

gdevenyi:
Sorry I will get back to you pointing to the right place. Working from home + toddler...

cookpa:
Thanks, no rush.

gdevenyi:
My memory is bad, my control of installing libraries is here and works: https://github.com/ANTsX/ANTs/blob/master/ANTS.cmake#L194-L198

It only applies to ITK/VTK libs as part of the superbuild.

It would be nice to have an option that
1. Skips building the programs if this is enabled.
2. Doesn't install the ANTs libs if ANTS_INSTALL_PROGRAMS_ONLY

cookpa:
Thanks everyone. An install binaries only option would also be handy, if anyone knows how to do that.

PR #978, opened 2020-03-23
N4 verbosity

muschellij2:
I have to do some testing to see if this will compile, but @ntustison - I think this could be a start to see where https://github.com/ANTsX/ANTsRCore/issues/109 is happening

### DIFF
--- a/Examples/N4BiasFieldCorrection.cxx
+++ b/Examples/N4BiasFieldCorrection.cxx
@@ -71,12 +71,6 @@ int N4( itk::ants::CommandLineParser *parser )
 {
   using RealType = float;
 
-  using ImageType = itk::Image<RealType, ImageDimension>;
-  typename ImageType::Pointer inputImage = nullptr;
-
-  using MaskImageType = itk::Image<RealType, ImageDimension>;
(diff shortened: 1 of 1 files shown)


muschellij2:
I think this should be good to go as well.

ntustison:
Why are we pushing this to the main repository?  We've already determined the probable location of where the seg fault is occurring.  The problem, as I've discussed https://github.com/ANTsX/ANTsRCore/issues/109, is my not being able to reproduce the error.

muschellij2:
Agreed, but I still get segfaults as well as the GitHub actions build.  I think a little bit more verbosity can't hurt, but I'll defer to you to accept or reject.

ntustison:
Thanks @muschellij2 .  The proposed verbose statements were meant to root out the location of the segfault.  I think they're excessive for the general use case (e.g., we don't need inform the user that an image type is being created).  I think to move this forward you, or somebody on your team, who can reproduce the error needs to debug this.  I've already pointed to a possible location in the code but, as I can't reproduce the segfault,  I'm not going to debug this and I doubt @stnava will either.

PR #980, opened 2020-03-24
ENH: Sharpening options for AverageImages and ImageMath

cookpa:
Not changing the template scripts for now, but here's some new features:

ImageMath UnsharpMask command, which provides access to this filter

https://itk.org/Doxygen/html/classitk_1_1UnsharpMaskImageFilter.html

and its main parameters: amount (default 0.5, useful range 0-2), radius (default 1mm) and threshold (default 0). The last option lets you specify radius in spacing units or not, like SmoothImage.

AverageImages now has three options:

0 - just get the mean in each voxel
1 - mean of each voxel, but input volumes are divided by the mean within the volume, then Laplacian sharpening is applied to the result
2 - mean of each voxel, but input volumes are divided by the mean within the volume. No sharpening

This allows users to sharpen using their choice of method with ImageMath, or have backwards compatibility by choosing option 0 or 1.


gdevenyi:
Looking forward to the test results from the modification to the pipeline :+1:

PR #1012, opened 2020-05-29
COMP: Fix compilation errors with latest cmake version

laurennlam:
With the latests cmake versions, in the GUI, the generator plateform is not included into the CMAKE_GENERATOR. The default value of the generator plateforme (if it's not set by the user from the GUI), it's x84. Then, if the user sets x64, the compilation failed. All external project need to know the correct generator plateform.


gdevenyi:
@laurennlam can you please provide a complete report of what exactly this is trying to fix?
- cmake version
- OS
- complete logs

Thanks.

laurennlam:
I used CMake 3.16.2 on Windows (and I only used the CMake GUI, not command line)
Here is the error I had on the master branch : 

> fatal error LNK1112: computer type module 'x86' is in conflict with target computer 'x64'

When I launched the superbuild, it tried to build ITK (and all other external project) in x86 but I explicitly define x64 on CMake GUI.

cookpa:
Thanks @laurennlam

PR #1045, opened 2020-07-14
WIP:  Add code of conduct.

ntustison:
Still need to specify reporting procedures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@
 =========================================================
 
 [![Build Status](https://travis-ci.org/ANTsX/ANTs.svg?branch=master)](https://travis-ci.org/ANTsX/ANTs)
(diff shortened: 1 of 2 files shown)


cookpa:
Re: reporting. I'm happy to help out as needed. As a collaborator, I can lock threads that get out of hand, and remove content that is objectionable. Project admins (Nick and Brian) can give people timeouts or even perma-bans from ANTsX.

https://docs.github.com/en/github/building-a-strong-community/blocking-a-user-from-your-organization

As far as I know, there's no way to privately message people unless they have put an email in their profile. So moderation would potentially have to be in the threads.

ntustison:
Thanks @cookpa .  Much appreciated.  Once I get back from vacation, I can work a bit more to organize the reporting.

PR #1051, opened 2020-07-22
Use BinShrinkFilter instead of ShrinkFilter for N4 Downampling

gdevenyi:
Through experiments with N4BiasFieldCorrection, I've found that with noisy data N4 can be weirdly unstable to the underlying noise field of the data, through choices such as exactly how much padding the image gets. I believe this is due to the use of the ShrinkImage filter, which simply takes every Nth voxel to achieve downsampling. If we instead use the BinShrinkImage filter when downsampling, we reduce the noise and stabilize estimates for noisy data by performing a smoothing.

This isn't the only solution to this problem, we could explicitly smooth with a gaussian before downsample, but this the fastest drop-in replacement.

Ref: https://hdl.handle.net/10380/3450

### DIFF
--- a/Examples/N4BiasFieldCorrection.cxx
+++ b/Examples/N4BiasFieldCorrection.cxx
@@ -14,6 +14,7 @@
 #include "itkLabelStatisticsImageFilter.h"
 #include "itkN4BiasFieldCorrectionImageFilter.h"
 #include "itkShrinkImageFilter.h"
+#include "itkBinShrinkImageFilter.h"
 #include "itkTimeProbe.h"
 
 #include <string>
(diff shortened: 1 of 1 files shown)


cookpa:
Very interesting. How do you think the padding affects the results? I can imagine it might matter with a tight FOV, because the padding might introduce space around the edge of the imaged object, allowing the spline to fit better. We see the same with registration - the warp field is constrained to be zero at the edges of the image volume, and you want this to be some distance from the brain or whatever it is you're registering.

gdevenyi:
What I was more thinking was going on was that the padding would offset the shrink slices being selected, and so the exact voxels which end up being selected for consideration by N4 end up changing. Doing a binning during subsampling makes sure the local information is mixed in prior to downsampling, making this effect far less pronounced.

ntustison:
Hi @gdevenyi ,

I really appreciate these efforts.  However, next time I would recommend bringing up such issues in the forum before creating a pull request.  Many of us have been developing and using these algorithms for quite some time and these sorts of issues might have been discussed informally and/or offline and already settled and so creating a pull request might be a wasted effort on your part.  

As you mention, there are several possible alternatives that one can "drop-in" to N4 (bin shrink filter is probably the least optimal with a better alternative being to use our adaptive noise filter) but they are all essentially a preprocessing step, _hidden from the user_, to deal with a noisy input image.  If such an image is causing such problems with the current implementation, then good practices dictate that the noise be dealt with explicitly outside, and probably prior to, a bias correction algorithm.  Considering this and the fact that these are probably peripheral cases and the fact that I'm reluctant to change core algorithms, generally, because you never know how that will effect our own pipelines as well as those of other users, I'm disinclined to merge something like this.

gdevenyi:
Okay, I'll consider ANTs maintenance-only and write my own workarounds. Thanks for the feedback @ntustison

PR #1062, opened 2020-08-04
ENH: Use GNUInstallDirs CMake module

gdevenyi:
This is @cookpa's work. Given it fixes some outstanding bugs #872 we should merge.

Trying to enable installation to the correct lib dir for each platform.

This just needs the SuperBuild bits still adjusted, I will provide feedback.


gdevenyi:
I believe this line: https://github.com/ANTsX/ANTs/blob/master/ANTS.cmake#L199

Needs to have "lib" changed to ${CMAKE_INSTALL_LIBDIR}

cookpa:
I think I've set all the variables correctly, but CMAKE_INSTALL_LIBDIR does not get propagated to the ANTS-build. A workaround is to go to ANTS-build and then

```
cmake -D CMAKE_INSTALL_LIBDIR=lib64 .
make install
```

I'm trying to figure out how to make ANTS use the Superbuild value

gdevenyi:
This is probably related to "make install" not working properly at the superbuild level, but rather needs to happen inside ANTS after the final build.

cookpa:
I think I may have finally cracked it. Things were going wrong because when `CMAKE_INSTALL_LIBDIR` was `lib64`, ITK libs were going to `staging/lib` but VTK libs were going to `staging/lib64`. Now both stage to `lib` but then install to `${CMAKE_INSTALL_LIBDIR}`.

I would be grateful if anyone could test this on a non-Mac OS system.

PR #1063, opened 2020-08-08
Verbose registration

muschellij2:
Should show the seed values now in cout, related to https://github.com/ANTsX/ANTsR/issues/210


gdevenyi:
If we merge this lets make sure we do a squash so we don't see all these no-op commits.

PR #1064, opened 2020-08-10
Verbose output squashed commit

muschellij2:
Squashed verbosity for printing out.


gdevenyi:
:+1: squashing is actually possible next to the merge button in the PR as well but thanks for the update!

hjmjohnson:
This is very old issue and does not seem have support for integration.  If it is still desired, please reopen, rebase, and add a commit message with the rationale why this is beneficial.

PR #1078, opened 2020-09-07
BUG:  Fix expansion factors.

ntustison:
https://github.com/ANTsX/ANTs/issues/1076


cookpa:
Looks reasonable to me. Warp dimensions are correct and deformations resemble those before the bug in my test.

ntustison:
Awesome.  Thanks @cookpa .

PR #1115, opened 2020-11-19
ENH:  Reconcile defaults values in KK.

ntustison:


ntustison:
Hey @stnava and @cookpa , I figure we could reconcile with what was used in the cortical thickness paper.

cookpa:
Thanks - seems like a good idea to have them consistent in different ANTsX places.

PR #1125, opened 2020-12-19
COMP: CMake 3.19 has problems with ITK, set version to 3.18 in Docker

cookpa:
There's an issues with CMake 3.19 that breaks compilation of ITK. ITK has downgraded to 3.18.3 until it's fixed. See

https://gitlab.kitware.com/cmake/cmake/-/issues/21529


ntustison:
Thanks @cookpa .

PR #1129, opened 2021-01-14
ENH: Add option for SyN gradient step in scripts

cookpa:
Eg, `-g 0.2`. The default is unchanged (0.1). 

I've found that 0.1 can sometimes be too conservative at the SyN stage.


ntustison:
Thanks @cookpa

PR #1145, opened 2021-02-25
ENH: Added bash script for ANTs tissue segmentation

aghayoor:
ANTs Atropos and antsAtroposN4.sh need all the priors already warped to the subject space before the segmentation is run. This developed script combines the registration step with antsAtroposN4 run.

This script runs:
- brain extraction
- template registration
- atropos segmentation
- (optional) register to template space
- QC (mosaic QC image)

This script was adopted and modified from antsCorticalThickness.sh,
so it follows similar command line convention.

This script was tested on several arbitrarily selected
subject/templates, and it ran successfully and generated the requested outputs.

### DIFF
--- a/Scripts/antsTissueSegmentation.sh
+++ b/Scripts/antsTissueSegmentation.sh
@@ -1,0 +1,1173 @@
+#!/bin/bash
+
+VERSION="0.0"
+
+# Check dependencies
+
+PROGRAM_DEPENDENCIES=( 'antsRegistration' 'antsApplyTransforms' 'N4BiasFieldCorrection' 'Atropos' )
+SCRIPTS_DEPENDENCIES=( 'antsBrainExtraction.sh' 'antsAtroposN4.sh' )
+
+for D in ${PROGRAM_DEPENDENCIES[@]};
+  do
+    if [[ ! -s ${ANTSPATH}/${D} ]];
+      then
+        echo "Error:  we can't find the $D program."
+        echo "Perhaps you need to \(re\)define \$ANTSPATH in your environment."
+        exit
+      fi
+  done
+
+for D in ${SCRIPT_DEPENDENCIES[@]};
+  do
+    if [[ ! -s ${ANTSPATH}/${D} ]];
+      then
+        echo "We can't find the $D script."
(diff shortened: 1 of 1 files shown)


aghayoor:
@ntustison : Hi Nick, please review this PR. Hope it helps the ants community.

cookpa:
Hi,

Does this differ from running individual stages of antsCorticalThickness.sh?

aghayoor:
@cookpa :
It generates less confusion. For example when you want to skip just one middle step of antsCorticalThickness.
Can you skip only step "5: DiReCT cortical thickness" and run all other steps (including QC) just by using command line? For this I was using a hacky solution to copy needed OutputNameACTStageNComplete.txt to the output directory.

However, this script is dedicated to tissue segmentation and can reliably be used in segmentation workflows.

ntustison:
Thanks @aghayoor .  I'll try to take a look tomorrow and follow up with any questions.

ntustison:
Is this only applicable to the scenario mentioned above where you want to perform up to the segmentation stage and then skip to quality control?  If so, this would seem to have significant redundancy where we could just edit ``antsCorticalThickness.sh`` to perform segmentation-based QC immediately after the segmentation stage.

aghayoor:
Yes, it basically removes the step "5: DiReCT cortical thickness" and trims the code and the command line so it is only dedicated to segmentation.

ntustison:
Thanks @aghayoor .  Is there a reason why we can't just incorporate the improvements proposed by this script into the antsCorticalThickness.sh script?

aghayoor:
Hi @ntustison ,
I think we can add developments to  antsCorticalThickness pipeline, so it can skip a specific stage.
However, it was my preference to make a dedicated pipeline for segmentation because each part of this pipeline needed to be validated.
We can also further develop this pipeline so it does not need the input priors in a specific order the way that antsCorticalThickness requires that.

ntustison:
I can appreciate needing each part of the segmentation pipeline to be validated but, correct me if I'm wrong, isn't the segmentation pipeline simply a subset of the cortical thickness pipeline?

hjmjohnson:
Re-open if there is a desire to modify to better integrate with ANTs

Issue #956, opened 2020-02-25
Fix sharpening artifacts in template construction


cookpa:
**Is your feature request related to a problem? Please describe.**

The Laplacian sharpening during template construction introduces an artifact near edges with large contrast gradients. In T1w templates this is most obvious as a bright fringe along the edge of the corpus callosum, adjacent to the lateral ventricles.

**Describe the solution you'd like**

A sharpening option that avoids this artifact, but still de-blurs the template acceptably.

**Describe alternatives you've considered**

There is an itkOptimalSharpeningFilter in ANTs/Utilities, but it has not been used in many years. Maybe one of the deconvolution filters in ITK could help?

**Additional context**

A template before and after the sharpening operation.

<img width="896" alt="average" src="https://user-images.githubusercontent.com/611507/75262906-3224e100-57bb-11ea-9079-b510f8247e52.png">
<img width="892" alt="sharpened" src="https://user-images.githubusercontent.com/611507/75262909-34873b00-57bb-11ea-8992-67a532b7d085.png">

cookpa:
Inspired to revisit this by @ins0mniac2, who when I mentioned sharpening, immediately recognized it as a known problem in photography: the "halo effect".

ntustison:
@cookpa --- I don't have a solution but I'd like to look more closely.  Can you post both those files?

cookpa:
https://upenn.box.com/s/a9c0cns71pi07vq6c3y2imma3klrfown

Apologies for the size, it's a high resolution template.

For the unsharpened version I did ImageMath Normalize, then AverageImages without normalization / sharpening. Then I called ImageMath Sharpen, which uses the same Laplacian filter as AverageImages.

ntustison:
Hey @cookpa ,

@stnava and I discussed this yesterday, specifically in the context of deep learning approaches.  I wanted to apply his [super resolution](https://github.com/ANTsXNet/MRISuperResolution) network to these images just to see the effects.  I'm attaching a screenshot in case you were interested.

[Superresolution.pdf](https://github.com/ANTsX/ANTs/files/4257323/Superresolution.pdf)

stnava:
yes - that's one option.  another is training a new network to infer the best "mean" anatomy from the set of images.   this would be like a deep learning version of JIF - which is yet another option.

ins0mniac2:
Meanwhile, can we get a user option to tune the sharpening ?

cookpa:
[...]

PR #1149, opened 2021-02-26
Sharpening options in template construction

cookpa:
This is mostly to address #956, but also adding a few other changes to improve performance.

Users may choose no sharpening, Laplacian (default), or unsharp mask.

Summary statistic: Made mean, median available in antsMultivariateTemplateConstruction.sh

Backwards compatibility is mostly preserved by leaving normalized mean (-a 1) and Laplacian sharpening (-A 1) as the defaults. I did slightly modify how sharpening is used in initial templates:

* Don't sharpen initial templates after population averaging or initial rigid alignment. These initial templates are often very messy and arguably should be smoothed, not sharpened

* In antsMultivariateTemplateConstruction2.sh, use median for initialization - again to help convergence and avoid outliers causing too much trouble.

Misc fixes:

* If -z is specified, ensure that the template exists and don't silently fall back to a population average.


cookpa:
I need to test ` antsMultivariateTemplateConstruction.sh` on another machine because I'm having trouble with it on my Mac.

gdevenyi:
Is it worth adding a median method to `AverageImages` so that the its available elsewhere and a bit cleaner?

cookpa:
Yeah, updating AverageImages would probably be good. I've been wanting to make an antsAverageImages using the updated CLI, but I have not had time to do that.

cookpa:
Anyway it looks like the sharpening options are working correctly in both templates, so I'm taking this out of draft mode.

PR #1151, opened 2021-03-02
DOC: Include version information and citation in docker images

cookpa:
Adding a label at build time to identify the source code version from an automated build using docker inspect.


cookpa:
I need to check the circleci output before merging this

ntustison:
Awesome.  Thanks, @cookpa

PR #1154, opened 2021-03-03
ENH: Set Docker container creation timezone

cookpa:
Trying again by moving the environment setting within the run block


cookpa:
This didn't work either. I'm going to leave it for now.

Issue #614, opened 2018-07-25
can't read initial transform ANTs0GenericAffine.mat


carolin31:
Hi,

I am trying to run antsApplyTransforms to apply the transformation parameters from a structural-to-template warping to functional data (bucket data set from afni, which contains several sub-bricks with each sub-brick giving beta weights or t-values for a specific regressor of interest). structural and functional data have been of course previously coregisterd to each other.
I used the following code:
antsApplyTransforms -d 4 -v 1 -r Zach_Template.nii -i stats.s201_REML.nii -o stats_warped_ -t ANTs1Warp.nii.gz -t ANTs0GenericAffine.mat

I get the following message:

Using double precision for computations.
Input scalar image: stats.s201_REML.nii
Reference image: Zach_Template.nii
Can't read initial transform ANTs0GenericAffine.mat

I found that somebody posted a similar problem on source forge:https://sourceforge.net/p/advants/discussion/840261/thread/c0485b2c/?limit=25

I still don t understand what causes the problem and how I can fix it?

Thank you in advance 
Carolin

ntustison:
Hi @carolin31 , Can you post the data so I can check it on my machine?

carolin31:
Hi Nick,

thank you very much for the quick response. The data is on Zion. Here is
the link:

/tmp/yassamri/R21/caro/201/ANTs


Let me know, if you should not be able to access them.

Thanks again!

Caro

ntustison:
I don't know what the problem is.  You have a couple things incorrect on your command line but they don't have anything to do with the error your getting.  However, fix them with the command below and tell me what you get.

```
$ antsApplyTransforms -d 3 -v 1 -r Zach_Template.nii -i stats.s201_REML.nii -o stats_warped.nii.gz -t ANTs1Warp.nii.gz -t ANTs0GenericAffine.mat
```

ntustison:
Hi @carolin31 ,

ITK nifti I/O reads the stats.s201_REML.nii image as a 3-D image with each voxel having 30+ components.  Note how this is different from a 4-D image where the fourth dimension is used to store the 30+ components.  The latter could be handled via the time-series option in ``antsApplyTransforms``.  Can you extract each of these components to separate scalar 3-D volumes?  

Nick

carolin31:
Hi Nick, 
so are you saying that I can not use antsApplyTransforms on this kind of dataset? And that I have to find a way to first transform the stats file into individual images?

[...]

PR #1191, opened 2021-05-13
ENH:  Accommodate AFNI stats images.

ntustison:
Revisiting [this issue](https://github.com/ANTsX/ANTs/issues/614).  Don't think the multi-channel works.  Need to explicitly read it in as a 5-D image.   @cookpa , can you check this if you have some time?


cookpa:
I think this makes sense. To confuse matters, tensors in NIFTI-1 are also 5D images with a specific intent code. I can add to the Wiki to clarify further.

ntustison:
Yeah, that would be great.   I figured you'd be aware of any relevant Nifti details.

Issue #1189, opened 2021-05-10
Beginnings of a default set up for reproducible registration in antsx


stnava:
**goal**

Provide a generally agreed upon and easily accessible set of registration parameters that support reproducible results with multi-threading

**proposed solution**

0. set random seed to 1 (though still allow user to  control this)
1.  Use GC metric for translation/rigid/affine
2.  Use CC metric for deformable ( radius 2 for quick settings )

**example implementation**

See ANTsPy example (somewhat tested): [here](https://github.com/ANTsX/ANTsPy/commit/82b662c9ff0b574da8a21be7d2a2fb14574afc4c)

**issues**

* We don’t know how well this setup will perform for “in the wild” data

* `GC` is not as general as `MI`

**note**

This function can help to some extent with the limitations of correlation metrics:

```
def rank_intensity( x, method='max' ):
    fim = ants.get_mask( x )
    fir = rankdata( (x*fim).numpy(), method=method )
    fir = fir - 1
    fir = fir.reshape( fi.shape )
    rimg = ants.from_numpy( fir.astype(float)  )
    rimg = ants.iMath(rimg,"Normalize")
    return( ants.copy_image_info( x, rimg ) )
```

It transforms intensities to reduce impact of outlier intensities or the effects of “unusual” histogram shapes.

For T1, the transformed image still looks “good” … not sure about other modalities.

**what would need to be done**

* standardize parameters if the ones I've proposed are not agreeable
* propagate the same approach to .sh scripts and `ANTsR`
* verify results are reproducible across runs and n-threads for each implementation
* potentially implement a specific version for MI as well where multi-modal registration is concerned
* alternatively, implement a reproducible variant of the MI metric --- doable by truncating precision but this may be a fragile solution and some other approach (unknown) may be better

any thoughts appreciated.

stnava:
my preliminary relatively large scale (2000 images) testing looks ok for these parameters .... will look more soon but perhaps it would be worth putting a beta version into ANTsR, at least, if not ANTs.

ntustison:
Okay, I'll put that on my to-do list to get to quickly.  Will let you know when I finish.

ntustison:
https://github.com/ANTsX/ANTsRCore/pull/132

and

https://github.com/ANTsX/ANTs/pull/1190

stnava:
nice - I also added rank_intensity/rankIntensity to antspy/R .... remains to be seen how useful they are.

cookpa:
[...]

PR #1196, opened 2021-05-17
ENH: Allow user to set random seed in repro mode

cookpa:
These changes allow the user to control the random seed in repro mode (as suggested in #1189).

Also expanded upon the usage a bit


ntustison:
Thanks @cookpa

PR #1218, opened 2021-08-06
COMP: Use modern ITK paradigms

hjmjohnson:
Future versions of ITK will deprecate and remove outdated coding
paradigms.  Replace deprecated code with modern coding recommendations.

Use official ITK5.2.0 release with improved backward/forward
compatibility syntax.


ntustison:
This is great, @hjmjohnson .  Thank you very much.

PR #1227, opened 2021-08-18
Template initialization to improve convergence

### LINKED ISSUE #1216: Templates not sized according to the largest input image
As part of some changes described here

https://github.com/ANTsX/ANTs/pull/1149

I changed the population average initial template to be median instead of mean

https://github.com/ANTsX/ANTs/blob/61e9501f639663cdd3ae08b8840a85c7b805a8d3/Scripts/antsMultivariateTemplateConstruction2.sh#L1012

But this has the undesired effect of not making the initial template the size of the largest input image.

Possible solutions:

- Compute median in AverageImages. This would be ideal from a scripting perspective as they would no longer need ImageSetStatistics at all

- Resize output in ImageSetStatistics, matching behavior of AverageImages

- Find some better way to initialize templates

- Revert to using mean. I'd like to avoid this because the initial template can be pretty bad if the input images are not roughly aligned.

cookpa:
Fixes #1216. In addition to enforcing the use of AverageImages in population average templates, I made a few tweaks to help convergence:

* Use @gdevenyi 's idea to do a COM alignment of each image to the initial average
* Don't sharpen the population average, resulting in a smoother initial template 
* Also don't sharpen after the initial rigid alignment

After the population average + initial rigid stage, the statistic (mean, normalized mean, median) and sharpening options follow whatever the users specify on the command line.
 
In order to make the process more transparent to users, I added an output folder `intermediateTemplates/`, which contains each template and, if it exists, the template shape update. This will let users visually inspect their templates at every stage, and the update warps can be interrogated (eg, for magnitude or volume change) to assess convergence. 

Currently, the only way to retain intermediate output is to back up everything, which can take up a lot of disk space. 

Additionally, I made a few fixes to documentation and error checking, including:

* Check number of shrink factor levels is equal to number of iteration levels (2.sh only)
* Check the expected number of warped output images exist for each shape update. This prevents a quiet error when one or more registrations fail for whatever reason, and the template gets built from a subset of the input.

### DIFF
--- a/Scripts/antsMultivariateTemplateConstruction.sh
+++ b/Scripts/antsMultivariateTemplateConstruction.sh
@@ -123,12 +123,23 @@ Optional arguments:
 
      -p:  Commands to prepend to job scripts (e.g., change into appropriate directory, set paths, etc)
 
-     -r:  Do rigid-body registration of inputs before creating template (default 0) -- 0 == off 1 == on. Only useful when
-          you do not have an initial template
+     -r:  Do rigid-body registration of inputs to the initial template, before doing the main
+          pairwise registration. 0 == off 1 == on (default 0). If you are trying to refine or update
+          an existing template, you would use '-r 0'.
+          Rigid initialization is useful when you do not have an initial template, or you want to use
+          a single image as a reference for rigid alignment only. For example,
+            "-z tpl-MNI152NLin2009cAsym_res-01_T1w.nii.gz -y 0 -r 1"
+          will rigidly align the inputs to the MNI template, and then use their average to begin the
+          template building process.
 
-     -s:  Type of similarity metric used for registration.
+     -s:  Type of similarity metric used for nonlinear registration (affine is always MI). Default = CC.
+          Options are case sensitive.
+             CC  : Cross-correlation
+             MI  : Mutual information
+             MSQ : Mean squared differences
+             PR  : CC after subtraction of local mean from the image (deprecated)
 
-     -t:  Type of transformation model used for registration. Supported options are case sensitive.
+     -t:  Type of transformation model used for nonlinear registration. Options are case sensitive.
              GR             : Greedy SyN (default for scalar data)
              GR_Constrained : Greedy SyN with regularization on the total deformation (default for time series)
              EL             : Elastic
@@ -144,10 +155,15 @@ Optional arguments:
           component of the affine transform will not be used to update the template. If your
           template drifts in translation or orientation try -y 0.
 
-     -z:  Use this this volume as the target of all inputs. When not used, the script
-          will create an unbiased starting point by averaging all inputs. Use the full path!
+     -z:  Use this this volume as the target of all inputs. When not used, the script will create an unbiased
+          starting point by averaging all inputs, then aligning the center of mass of all inputs to that of
+          the initial average. If you do not use -z, it is recommended to use "-r 1". Use the full path.
+          For multiple modalities, specify -z modality1.nii.gz -z modality2.nii.gz ...
+          in the same modality order as the input images.
 
-     -b:  Boolean for saving iteration output to directories (default = 0).
+     -b:  Boolean for saving full iteration output to directories (default = 0). If 1, images and warps
+          are saved for each pairwise registration at each iteration. Otherwise, only templates and the shape
+          update warps are saved.
 
 Example:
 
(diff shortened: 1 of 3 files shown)


hjmjohnson (approved):
These look like good changes.  Thank you for the style fixes cleaning of eol spaces.

The temporary directory will be great for debugging.

gdevenyi:
Looks great! My only suggestion is to perhaps improve the documentation a bit to clarify how the different modes work with this new COM modification.

Scripts/antsMultivariateTemplateConstruction.sh
```diff
@@ -837,13 +859,34 @@ for (( i = 0; i < $NUMBEROFMODALITIES; i++ ))
...
+            IMGbase=`basename ${CURRENTIMAGESET[$j]}`
+            BASENAME=` echo ${IMGbase} | cut -d '.' -f 1 `
+            COM="${OUTPUT_DIR}/initialCOM${i}_${j}_${IMGbase}"
+            COMTRANSFORM="${OUTPUT_DIR}/initialCOM${i}_${j}_${BASENAME}.mat"
+            antsAI -d 3 --convergence 0 --verbose 1 -m Mattes[${TEMPLATES[$i]},${CURRENTIMAGESET[$j]},32,None] -o ${COMTRANSFORM} -t AlignCentersOfMass
+            antsApplyTransforms -d 3 -r ${TEMPLATES[$i]} -i ${CURRENTIMAGESET[$j]} -t ${COMTRANSFORM} -o ${COM} --verbose
+            rm -f $COMTRANSFORM
+            IMAGECOMSET[${#IMAGECOMSET[@]}]=$COM
+          done
+          ${ANTSPATH}/AverageImages $DIM ${TEMPLATES[$i]} 2 ${IMAGECOMSET[@]}
```
gdevenyi, line 875: Looks here like we can't initialize with anything other than a normalized mean.

It seems to me that since we're resampling into a common space we could properly use the median method here, if the user wants it.

I think this should correspond in general to the average type the user requests.

cookpa: It could; it's just a little more work to do in n.sh rather than n2.sh because of the lack of a `summarizeimageset` function.

Scripts/antsMultivariateTemplateConstruction.sh
```diff
@@ -837,13 +859,34 @@ for (( i = 0; i < $NUMBEROFMODALITIES; i++ ))
         echo "--------------------------------------------------------------------------------------"
         # Normalize but don't sharpen at this stage
         ${ANTSPATH}/AverageImages $DIM ${TEMPLATES[$i]} 2 ${CURRENTIMAGESET[@]}
+        # Quickly align COM of input images to average, and then recompute average
+        IMAGECOMSET=()
+        for (( j = 0; j < ${#CURRENTIMAGESET[@]}; j+=1 ))
```
gdevenyi, line 864: One other implementation note here, I found for best performance I often had to do two rounds of this.

Scripts/antsMultivariateTemplateConstruction.sh
cookpa: OK I made this possible now. I think median is less useful now that we are making the initial average better, but users can choose.

Scripts/antsMultivariateTemplateConstruction.sh
cookpa: Do you have an example of this? I can't see how it would help except to shift the location of the image set inside the bounding box. Which may be desirable (or not if it shifts the wrong way)

gdevenyi: After checking things over, my tests show I don't actually need two rounds, however, padding/recropping the template is important as otherwise you can get a non-centered FOV.

Compare, left no pad/recrop vs right:
https://imgur.com/a/ded9IPL

Pad/recrop code:
https://github.com/CoBrALab/optimized_antsMultivariateTemplateConstruction/blob/master/modelbuild.sh#L486-L490

cookpa: I agree that's a good idea, but this whole PR started because a user wanted to control the size of their template image. Also, unless `-y 0` is used, the template will re-center itself to minimize the average translation.

cookpa:
I'm going to merge this now. Thanks for the reviews and suggestions

gdevenyi (approved):

ntustison:
Thanks for all this work, @cookpa and others @gdevenyi @hjmjohnson.

PR #1241, opened 2021-09-23
ENH: Version info from a config file for releases

cookpa:
Did a bunch of refactoring of version handling, consolidating the work
into ProjectSourceVersion.cmake (handling the git side and setting
_GIT_VERSION_MAJOR etc) and Version.cmake (deciding the final form of
the version info).
    
The version string should now match git-describe more closely. The old
behavior was to look for commits since Version.cmake was changed, but
since that is entirely auto-populated, it is not a very useful metric.
Now use post-X, where X is the number of commits since the last
release.
    
For releases, we need to provide a file
ANTs/CMake/ReleaseSourceVersionVars.cmake
which should contain, eg:
    
set( _GIT_VERSION_MAJOR "2" )
set( _GIT_VERSION_MINOR "3" )
set( _GIT_VERSION_PATCH "6" )

### DIFF
--- a/ANTS.cmake
+++ b/ANTS.cmake
@@ -31,18 +31,10 @@ if( NOT ${PROJECT_NAME}_BUILD_DISTRIBUTE AND NOT ${PROJECT_NAME}_VERSION_HASH ST
   set(${PROJECT_NAME}_VERSION "${${PROJECT_NAME}_VERSION}-g${${PROJECT_NAME}_VERSION_HASH}")
 endif()
 
-# When building from a snapshot, the git version information does not get populated 
-# Allow the user to specify a hash or version tag on the command line
-if( ${PROJECT_NAME}_VERSION_HASH STREQUAL "GITDIR-NOTFOUND" AND NOT "${ANTS_SNAPSHOT_VERSION}" STREQUAL "" )
-  set(${PROJECT_NAME}_VERSION_MAJOR 0)
-  set(${PROJECT_NAME}_VERSION_MINOR 0)
-  set(${PROJECT_NAME}_VERSION_PATCH 0)
-  set(${PROJECT_NAME}_VERSION_TWEAK 0)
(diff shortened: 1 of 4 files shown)


cookpa:
I've removed the CMake option for users to include "snapshot-[hash]". Now users can specify version information at build time the same way as a build from git does.

cookpa:
@ghisvail, would this work for your purposes? I would add the file `ANTs/CMake/ReleaseSourceVersionVars.cmake` to a tarball and attach it to the release. Eventually I will figure out how to automate this. 

Alternatively, you could download Github's source tarball and use a script to write the version number into `ANTs/CMake/ReleaseSourceVersionVars.cmake`.

cookpa:
One downside to the CircleCI version file approach is that there would be two copies of the source for each release: Github's auto-generated archives of the repository, and one attachment that would contain the version information.

To get around this, we could make a releases branch that contains the `ANTs/CMake/ReleaseSourceVersionVars.cmake`, and release from there.

A third option is to revert to hard-coded numbers in `Version.cmake` on master. Undoing changes made by @gdevenyi some time ago.

ghisvail:
@cookpa our requirements for conda-forge is an appropriately versioned *release tarball* for ANTs.

> there would be two copies of the source for each release [...]

That does not sound very appealing, but ymmv.

> A third option is to revert to hard-coded numbers in `Version.cmake` on master.

Like pretty much everyone else does.

gdevenyi:
I can say that in minc-toolkit we ran into issues with auto-generated tarballs, as we use git submodules and the tarball generation doesn't pull them, so we had to make our own, since we had to anyways, modifying the version was less of an issue.

Similarly, with my qbatch project we also hard-code the version in setup.py and update it as the last commit before tagging a release (which Travis then autopublishes to pip)

cookpa:
Thanks for this feedback. I think I'm going to try a hard-coded version in the repo, with script that does the tags and updates the hard-coded versions.

PR #1243, opened 2021-09-28
Added support for IBM POWER8/9 architectures

pantaray:
Hello ANTs team!

We're running IBM POWER E880 servers in our HPC cluster infrastructure and I built ANTs for that architecture. Since building `libpng` on POWER is a little tedious I basically just added the option to use the system `libpng` for ITK and set compiler flags depending on the build target. Maybe these changes are useful for others who want to run ANTs on similar hardware.

All the best from Frankfurt,
Stefan


cookpa (approved):
Looks good to me

cookpa:
Thanks @pantaray

PR #1246, opened 2021-09-30
ENH: Version info in repo

cookpa:
ENH: Encode version information in repo
    
This avoids the placeholder version name 0.0.0.0 for tarball builds, which causes problems with external builds.
    
`Version.cmake` now contains hard-coded version numbers that are used
for builds (from git or tarball source). For releases, a variable
is set so that binaries have version vA.B.C. For other commits (not a
tagged release), the version is vA.B.C.dev. Builds from git repos
still contain the additional information about the number of commits
post-tag, and the short hash.
    
To make a release, first check out an up to date copy of master.
Then run `Utilities/tagRelease.pl <tag>`. This will update `Version.cmake`,
commit + push the changes, tag that commit, then change `Version.cmake`
back into dev mode, and commit + push the result. 
 
This means:

  * Version info for git users will be similar to before. I added more chars to the abbreviated hash to conform to git convention
  * When building a release, users just need to download the source tarball from the releases page, and the binaries will have version "vX.Y.Z"
  * Git users can also use 'cmake -DANTS_BUILD_DISTRIBUTE=TRUE .` before `make install`, to get "vX.Y.Z".
  * When building HEAD not on a tagged commit, the version info in a source tarball has "vX.Y.Z.dev", so we know it was not the actual release code, but can localize it somewhat.

This is an alternative approach to the one proposed in #1241, which required a separate file to be bundled into releases. Fixes #1241


gdevenyi:
> tag that commit, then change Version.cmake back into dev mode, and commit + push the result.

I think it's possible for the cmake/git code to detect if you're "on" a tag or not, such that you could avoid this dance and just append dev when you're not coincident with a tag>

cookpa:
> > tag that commit, then change Version.cmake back into dev mode, and commit + push the result.
> 
> I think it's possible for the cmake/git code to detect if you're "on" a tag or not, such that you could avoid this dance and just append dev when you're not coincident with a tag>

Git can detect that, and does in the code. But tarball users won't be able to use that. 

I just want to avoid people saying "I have v2.3.5" when they could have any of a few dozen commits post 2.3.5

PR #1247, opened 2021-10-03
BUG - antsIntermodalityIntrasubject.sh - fixed a typo and incorrect running summary

nirso:
Hi!

I've been running antsIntermodalityIntrasubject.sh and noticed that even though I am providing the brain mask, it is not included in the running "summary". While confusing at first, I then found the issue and the wrong parameter is being used in the summary.

You can see the differences here.
Before:
Snip20211003_12

After:
Snip20211003_13

Also, I fixed a small typo.

Thanks!


cookpa:
Thanks @nirso

PR #1265, opened 2021-11-04
BUG: Lesions in background or surrounded by darker tissue could cause…

### LINKED ISSUE #1258: LesionFilling seems not to save the resultant image.
I was running LesionFilling command on Linux VM (Xubuntu VBox). 
It is a very simple and straightforward script, but I got an error message shown below:
"""
Number of lesions: 174
./WML_filling_counting_labeling_volume-recording_Data_since_2019.sh: line 18: 21602 Floating point exception(core dumped) ${ANTSPATH}/LesionFilling 3 ./${i}-pre-denoised-T1W_0p5.nii.gz ./${i}-FLAIR_WML.nii.gz ./${i}-lesion_filled_T1W_0p5.nii.gz
'
"""
And the result (supposed to be a lesion-filled T1 image) was not saved.

I would like to know what may cause 'Floatingpoint exception(core dumped)'?
Are there any conditions for input data? (nii.gz or nii etc.)

Thank you,
SC

cookpa:
… a divide by zero

Fix this by checking lesion clusters for non-zero intensity. Also, for
very small clusters, dilate by 2 voxels instead of 1, to improve the
chance of finding suitable neighborhood voxels.

If lesions cannot be filled, they are skipped and a warning is printed

Fixes #1258

### DIFF
--- a/Examples/LesionFilling.cxx
+++ b/Examples/LesionFilling.cxx
@@ -15,6 +15,7 @@
 #include "itkBinaryBallStructuringElement.h"
 #include "itkMaskImageFilter.h"
 #include "itkConnectedComponentImageFilter.h"
+#include "itkRelabelComponentImageFilter.h"
 #include "itkBinaryThresholdImageFilter.h"
 #include "itkImageRegionIterator.h"
 //LesionFilling dimension t1.nii.gz lesionmask output.nii.gz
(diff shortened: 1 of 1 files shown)


cookpa:
@ntustison @dorianps this is mostly a bug fix but I did add extra dilation for small lesions. The problem encountered by a user was that for small lesions near CSF, the immediate neighborhood may contain 0 valid (ie, brighter) voxels. So I added extra dilation in that case to have a slightly larger pool of voxels to choose from. Does that sound OK?

ntustison:
Thanks @cookpa .  I would have to defer to @dorianps regarding this change.

dorianps:
Unfortunately I have not used LesionFilling that much. I created a separate approach based on JIF and a number of T1w from controls (starting with a symmetric left right flip to fill first with the subject itself). So, I can't say if this is good or not. From what you describe @cookpa  sounds ok.

cookpa:
OK thanks!

PR #1290, opened 2022-01-21
ENH: Use a bigger VM to attempt to stop CircleCI timeouts

cookpa:
CircleCI jobs timeout after one hour. The recent ITK upgrades push the build time over the limit, but perhaps a larger resource will be faster.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,7 @@ jobs:
   build_docker:
     docker:
       - image: docker:18.06.1-ce-git
+    resource_class: large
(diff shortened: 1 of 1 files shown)


cookpa:
Actually it looks like the ITK update didn't cause this, it just coincided with CircleCI's decision to start limiting free jobs to one hour.

gdevenyi:
What are the limits on Github Actions? Is it time to try it out?

cookpa:
Yes that could be an option

https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration

PR #1305, opened 2022-02-05
COMP: Update ITK and VTK to avoid compiler warnings

hjmjohnson:
Older versions of ITK and VTK introduced compiler warnings
with new more pedantic compilers.  VTK warned of duplicate
symbols, and ITK warned of empty statements.

Code built and tests passed:

100% tests passed, 0 tests failed out of 130


ntustison:
Awesome.  Again, thanks @hjmjohnson

Issue #1315, opened 2022-03-07
Cannot read HDF5 transforms created with antsApplyTransforms


cookpa:
**Describe the problem**

I'm trying to create HDF5 transforms from with `antsApplyTransforms`, but the files I create are not read successfully.

**To Reproduce**

```
antsApplyTransforms -d 3 \
  -i ${TEMPLATEFLOW_HOME}/tpl-MNI152NLin6Asym/tpl-MNI152NLin6Asym_res-01_atlas-Schaefer2018_desc-100Parcels7Networks_dseg.nii.gz \
  -t ${TEMPLATEFLOW_HOME}/tpl-MNI152NLin2009cAsym/tpl-MNI152NLin2009cAsym_from-MNI152NLin6Asym_mode-image_xfm.h5 \
  -r ${TEMPLATEFLOW_HOME}/tpl-MNI152NLin2009cAsym/tpl-MNI152NLin2009cAsym_res-01_desc-brain_T1w.nii.gz \
    -o mni2009cLabels.nii.gz --verbose 
```

This works successfully, using an existing .h5 file in templateflow. But if I try to make my own, it fails to read. For example:

```
antsApplyTransforms -d 3 \
-i ${TEMPLATEFLOW_HOME}/tpl-MNI152NLin6Asym/tpl-MNI152NLin6Asym_res-01_atlas-Schaefer2018_desc-100Parcels7Networks_dseg.nii.gz \
  -t ${TEMPLATEFLOW_HOME}/tpl-MNI152NLin2009cAsym/tpl-MNI152NLin2009cAsym_from-MNI152NLin6Asym_mode-image_xfm.h5 \
  -r ${TEMPLATEFLOW_HOME}/tpl-MNI152NLin2009cAsym/tpl-MNI152NLin2009cAsym_res-01_desc-brain_T1w.nii.gz \
  -o [ mni2009cFrom6.h5, 1 ]  --verbose 
```

appears to work but then

```
[...]

PR #1317, opened 2022-03-11
ENH: Allow writing HDF5 transforms as ITK transforms 

cookpa:
HDF5 vector images don't work with transform readers.

### DIFF
--- a/Examples/antsApplyTransforms.cxx
+++ b/Examples/antsApplyTransforms.cxx
@@ -18,6 +18,7 @@
 #include "itkMatrixOffsetTransformBase.h"
 #include "itkTransformFactory.h"
 #include "itkTransformFileReader.h"
+#include "itkTransformFileWriter.h"
 #include "itkTransformToDisplacementFieldFilter.h"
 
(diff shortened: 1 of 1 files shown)


cookpa:
Feels wrong to put a file type dependence into `antsApplyTransforms.cxx`, but it works.

The other option would be to continue writing H5 files as vector images, and modify [itkantsReadWriteTransform.h ](https://github.com/ANTsX/ANTs/blob/cdef12fc2381af8a0eb6302eb28cd3ddd6d84a80/Utilities/itkantsReadWriteTransform.h) so that it tries to read an HDF5 file as a displacement field first. This strategy would avoid any file type testing in `antsApplyTransforms`, but the resulting .h5 files still won't be ITK transforms. I guess we could add a separate program to do that.

cookpa:
Addressing #1315

ntustison:
Thanks @cookpa .  I think @gdevenyi has a good point---if anybody is using ANTs with hdf5 files, I imagine they're using them elsewhere as well.

cookpa:
Fixes #1315

Issue #83, opened 2014-04-27
inverse transform error with antsApplyTransformsToPoints


satra:
> re: inverse transform:

```
$ antsApplyTransformsToPoints -d 3 -i in.csv -o out.csv -t reg/transformInverseComposite.h5 
Input csv file: in.csv
Transform reader for reg/transformInverseComposite.h5 caught an ITK exception:

itk::ExceptionObject (0x7fd2dc0150d8)
Location: "unknown" 
File: /software/builds/ANTs/ITKv4/Modules/Core/Transform/include/itkRigid3DTransform.hxx
Line: 127
Description: itk::ERROR: Rigid3DTransform(0x7fd2dc013c50): Attempting to set a non-orthogonal rotation matrix

Can't read initial transform reg/transformInverseComposite.h5
```

satra:
@stnava  and @hjmjohnson - do we know if this bug has been handled in ITK? we are still seeing this error on applying the inverse transform with ants version: 2.1.0-1.c0c854

and from the looks of the itk mailing list, the only google references were from 2009/10.

Issue #82, opened 2014-04-27
antsApplyTransformationsToPoints and CSV file coordinates are an exercise in obfuscation


satra:
in trying to debug this issue in mindboggle: https://github.com/binarybottle/mindboggle/issues/39

i created a very simple test following brian's chicken example (https://github.com/stnava/chicken).

```
A - brain
B - 3 labels drawn on A (shares the same voxel to world mapping)
C - Template
T - Transform A -> C using antsRegistration
D - CSV from B via ImageMath

E - T(A)
F - T(B)
G - T^{-1}(D) via antsApplyTransformationsToPoints using an inverse transform
H - CSV from F via ImageMath

G ~ H (this matches - woohoo! but --- read on)
```

however, when i took the xyz coordinates from D and put it into my viewers, the LR and AP dimensions were flipped. i.e. the coordinates were not respecting the voxel to world mapping of A/B.

Points for discussion:
1. D should reflect the voxel to world mapping of A/B (in 2d this can be identity)
2. the fact antsApplyTransformationsToPoints requires the inverse transform from antsApplyTransformations should be noted or **preferably changed**. it doesn't make any sense to have two semantically similar operations (antsApplyTransformations and antsApplyTransformationsToPoints) to require inverse transforms.
3. antsApplyTransformationsToPoints does not like the inverse composite transform provided by antsRegistration (this should be a separate issue)

PR #1331, opened 2022-04-06
WIP: PERF: Performance acceleration via GPU-accelerated FFT's

thewtex:
Further discussion today at the ANTs Get Your Brain Straight BoF:

  https://insightsoftwareconsortium.github.io/GetYourBrainStraight/HCK01_2022_Virtual/


ANTS.cmake
```diff
@@ -38,7 +38,64 @@ set(CPACK_RESOURCE_FILE_README "${CMAKE_CURRENT_SOURCE_DIR}/README.txt")
 message(STATUS "Building ${PROJECT_NAME} version \"${${PROJECT_NAME}_VERSION}\"")
 
 # Set up ITK
-find_package(ITK ${ITK_VERSION_ID} REQUIRED)
+find_package(ITK ${ITK_VERSION_ID} REQUIRED
```
cookpa, line 41: This block breaks the default SuperBuild for me. It's asking for a ITK components that aren't built by default:

VkFFTBackend
ITKFEM   
ITKVTK
ITKVtkGlue

I don't know about ITKFEM, but the others should be required only if the relevant build options are turned on.

gdevenyi: ITKFEM is because `Temporary/` has includes which use it.

VkFFTBackend is the code which supports this new feature.

ITKVTK and ITKVTK need to be wrapped so they only are required if the USE_VTK is on.

I'm working on this

gdevenyi: also `Examples/ConformalMapping.cxx` and `Examples/ANTSConformalMapping.cxx` but it appears these are not built, since ITKFEM module hasn't been enabled and the build still works.

cookpa: I was working on this as well, but I could not build with vkFFT support. It stops with this error

```
make                                                                                                                                                                   
[ 47%] Built target ITKv5
[ 52%] Performing forcebuild step for 'ANTS'
[ 58%] Performing build step for 'ANTS'
[  1%] Building CXX object Examples/CMakeFiles/antsUtilities.dir/antsRegistrationTemplateHeader.cxx.o
In file included from /Users/pcook/tmp/NOT_BACKED_UP/antsGPU/ANTs/Examples/antsRegistrationTemplateHeader.cxx:1:
In file included from /Users/pcook/tmp/NOT_BACKED_UP/antsGPU/ANTs/Examples/antsRegistrationTemplateHeader.h:14:
In file included from /Users/pcook/tmp/NOT_BACKED_UP/antsGPU/ANTs/Examples/itkantsRegistrationHelper.h:1472:
In file included from /Users/pcook/tmp/NOT_BACKED_UP/antsGPU/ANTs/Examples/itkantsRegistrationHelper.hxx:7:
In file included from /Users/pcook/tmp/NOT_BACKED_UP/antsGPU/build/staging/include/ITK-5.3/itkVkHalfHermitianToRealInverseFFTImageFilter.h:24:
/Users/pcook/tmp/NOT_BACKED_UP/antsGPU/build/staging/include/ITK-5.3/itkVkCommon.h:24:10: fatal error: 'vkFFT.h' file not found
#include "vkFFT.h"
         ^~~~~~~~~
1 error generated.
make[5]: *** [Examples/CMakeFiles/antsUtilities.dir/antsRegistrationTemplateHeader.cxx.o] Error 1
make[4]: *** [Examples/CMakeFiles/antsUtilities.dir/all] Error 2
make[3]: *** [all] Error 2
make[2]: *** [ANTS-prefix/src/ANTS-stamp/ANTS-build] Error 2
make[1]: *** [CMakeFiles/ANTS.dir/all] Error 2
make: *** [all] Error 2
```

Have you been able to compile with the vtFFT module on?

SuperBuild/External_ITKv5.cmake
```diff
@@ -80,6 +80,11 @@ if(NOT DEFINED ${extProjName}_DIR AND NOT ${USE_SYSTEM_${extProjName}})
         -DModule_ITKMINC:BOOL=ON
        )
   endif()
+  if(${extProjName}_BUILD_VKFFT_SUPPORT)
+    set(${proj}_MINC_ARGS
```
cookpa, line 84: I think this should be "${proj}_VKFFT_ARGS"

cookpa: Fixed

cookpa:
Oh, I guess edits aren't turned on, so my changes went to a local branch. 

https://github.com/ANTsX/ANTs/tree/perf-fft

Anyhow, I'm not sure if this PR is even still interesting, I was just curious to see if I could make it build. I got as far as it complaining about lack of Vulkan on my Mac.

hjmjohnson:
_Superseded — see the consolidated, current findings: https://github.com/ANTsX/ANTs/pull/1331#issuecomment-4736471293_

hjmjohnson:
_Superseded — see the consolidated, current findings: https://github.com/ANTsX/ANTs/pull/1331#issuecomment-4736471293_

hjmjohnson:
**VkFFT GPU-FFT findings for this PR — consolidated and current (Linux).** After fixing a per-call GPU resource leak and adding context + plan + buffer caching in `ITKVkFFTBackend` (PR #83), VkFFT is **compute-bound and stable**, and the **CUDA and OpenCL backends are essentially equal** (CUDA marginally faster at small sizes). On this discrete card, with a per-call host↔device copy, **PocketFFT (the ITK CPU default) is as fast or faster than VkFFT at 128³–256³**, so the PR **defaults to PocketFFT when no GPU is present** and routes to VkFFT only when a device is confirmed. Metal numbers are pending on a Mac (section below). _This comment supersedes my earlier benchmark comments on this thread._

<details>
<summary>Final steady-state numbers (RTX 6000 Ada, #83 @ 26e296d, 100 iters × 3, round-trip err 0.000)</summary>

3D single-precision real round-trip FFT, mean ms/call:

| size | PocketFFT (CPU) | VkFFT OpenCL | VkFFT CUDA |
|------|----------------:|-------------:|-----------:|
| 64³  | 0.65 | 0.68 | **0.60** |
| 128³ | **1.67** | 2.77 | 2.70 |
| 192³ | **3.62** | 8.20 | 8.20 |
| 256³ | **18.8** | 20.5 | 20.7 |

(FFTW ≈ 60 ms at 256³ — slower than PocketFFT, and pathological on non-power-of-2 sizes.)

Two takeaways:
- **CUDA is not slower than OpenCL** — they are within run-to-run noise, CUDA slightly ahead at 64³. VkFFT generates its own kernels for both backends, so once the plan is cached the compute is equivalent.
- For an **isolated** transform that copies host→device→host each call, **PocketFFT wins at these moderate sizes** — the GPU FFT compute is fast but the round-trip copy + launch overhead eats the gain. VkFFT's advantage appears for larger transforms and, crucially, when data stays **resident on the GPU** across operations (no per-call copy) — the realistic registration scenario, not exercised by this microbenchmark.

</details>

<details>
<summary>Optimization journey — why the numbers moved (256³)</summary>

| stage | VkFFT OpenCL | VkFFT CUDA | sustained reuse |
|-------|-------------:|-----------:|-----------------|
| Original (context + plan rebuilt every call) | 222 ms | 316 ms | **OOM after ~50 calls** |
| + GPU context caching | 37 ms | 147 ms | stable |
| + plan + per-shape buffer caching (#83) | **20.5 ms** | **20.7 ms** | stable |

The original code created (and leaked) a GPU context and recompiled the VkFFT plan on every `Update()`. The transient "OpenCL beats CUDA" seen mid-investigation was purely that recompile: the CUDA path re-ran **nvrtc** each call (~100 ms) while NVIDIA's OpenCL driver cached the compiled program (`~/.nv/ComputeCache`). Caching the context, the plan (`VkFFTApplication`), and the device buffers per shape — the canonical VkFFT usage — removes both the leak and the asymmetry. (Refs: [VkFFT/IWOCL 2023](https://dl.acm.org/doi/fullHtml/10.1145/3585341.3585357), [pyvkfft](https://pyvkfft.readthedocs.io/en/stable/) — both reuse/cache the plan.)

</details>

<details>
<summary>Correctness / GPU verification</summary>

- GPU execution confirmed via `nvidia-smi --query-compute-apps`; `compute-sanitizer --tool memcheck`: **0 errors**.
- Regression test `itkVkFFTRepeatedTransformLeakTest` (200 reused transforms): **fails at ~iter 54 with an out-of-memory abort on the pre-fix code, passes on the fixed code**.
- Round-trip error 0.000 across all sizes and shape changes, on both OpenCL and CUDA.

Upstream fixes from this work: ITK [InsightSoftwareConsortium/ITK#6465] (CMake `BUILD_INTERFACE` for build-tree system includes); `ITKVkFFTBackend` #82 (CUDA install/export, **merged**), #83 (context + plan + buffer caching, leak regression test, `VKFFT_BACKEND` ABI export — builds clean on Linux OpenCL and CUDA).

</details>

<details>
<summary>Metal (Apple GPU) — pending</summary>

_To be completed from a Mac using the VkFFT Metal backend (`VKFFT_BACKEND=5`). Results will be added here once gathered._

</details>

hjmjohnson:
_Superseded — see the consolidated, current findings: https://github.com/ANTsX/ANTs/pull/1331#issuecomment-4736471293_

hjmjohnson:
**Real-world check: GPU FFT does not accelerate ANTs registration smoothing at typical sizes.** ANTs already smooths with a recursive (IIR) Gaussian, which is ~25× faster than any FFT-based Gaussian, and VkFFT ≈ PocketFFT for the FFT path anyway. Details and the measurement below.

<details>
<summary>Smoothing benchmark — σ=3 voxels, mean ms/call, RTX 6000 Ada (CUDA for VkFFT)</summary>

| method | 128³ | 192³ | 256³ |
|--------|-----:|-----:|-----:|
| **`SmoothingRecursiveGaussianImageFilter`** (what registration uses) | **4.0** | **12.4** | **39.3** |
| `DiscreteGaussianImageFilter` (spatial) | 16.7 | 56.1 | 116.8 |
| `FFTDiscreteGaussianImageFilter` (PocketFFT) | 605 | 743 | 1010 |
| `FFTDiscreteGaussianImageFilter` (**VkFFT GPU**) | 601 | 740 | 1022 |

Gaussian smoothing is separable, so the recursive Gaussian is O(N) per axis and essentially unbeatable here; FFT convolution only pays off for large *non-separable* kernels. The GPU FFT (VkFFT) doesn't change this — it's ≈ PocketFFT for the FFT-Gaussian, and both are ~25× slower than the recursive filter the registration already uses.

</details>

<details>
<summary>Why the FFT-Gaussian factory is inert on the registration hot path</summary>

This PR registers `itkFFTDiscreteGaussianImageFilterFactory` (so `DiscreteGaussianImageFilter::New()` returns an FFT implementation) plus the VkFFT FFT factories. But the registration smoothing never constructs a `DiscreteGaussianImageFilter`:

- Per-level fixed/moving image smoothing → `itk::SmoothingRecursiveGaussianImageFilter` (`itkImageRegistrationMethodv4.hxx`).
- SyN/diffeomorphic field smoothing → `itk::GaussianOperator` neighborhood convolution (`itkGaussianSmoothingOnUpdateDisplacementFieldTransform`).

Neither is overridden by the FFT factory, so a standard `antsRegistration` runs identically with `ANTS_FFT_BACKEND=pocketfft` vs `vkfft` — the GPU path is never exercised by registration as written.

</details>

<details>
<summary>What the GPU FFT work did produce (kept separately)</summary>

The investigation still hardened `ITKVkFFTBackend` and ITK regardless of the ANTs outcome:
- A per-call GPU context/plan/buffer leak (OOM after ~50 transforms) fixed via context + plan + buffer caching; with that, VkFFT is compute-bound and CUDA ≈ OpenCL (see the consolidated findings comment above).
- Upstream: ITK InsightSoftwareConsortium/ITK#6465 (CMake `BUILD_INTERFACE`), ITKVkFFTBackend #82 (CUDA install/export, merged) and #83 (context+plan+buffer caching + leak regression test).

These are useful for any GPU-FFT-bound workload (large convolutions, spectral methods), just not for the separable Gaussian smoothing in registration.

</details>

<details>
<summary>Where a GPU could still help ANTs (out of scope here)</summary>

The registration cost is dominated by metric evaluation and field updates, not separable smoothing. A GPU win would more likely come from porting the metric / interpolation / field-exponentiation to the device and keeping image+field data resident (avoiding per-op host↔device copies), rather than FFT-accelerating a filter that recursive Gaussian already beats on the CPU.

</details>

hjmjohnson:
**Metal end-to-end registration numbers (Apple Silicon).** Ran `antsRegistration` (3D SyN, MeanSquares) with the VkFFT-Metal FFT backend vs the PocketFFT CPU default, built against ITK + ITKVkFFTBackend [PR #83](https://github.com/InsightSoftwareConsortium/ITKVkFFTBackend/pull/83) (`VKFFT_BACKEND=5`). **The two backends produce bit-identical results and essentially identical wall-clock** — the GPU-FFT win seen in isolation does not translate to a registration speedup, because FFT-based smoothing is not on the SyN hot path. This corroborates the PR defaulting to PocketFFT.

| volume | PocketFFT (mean) | VkFFT-Metal (mean) |
|--------|-----------------:|-------------------:|
| 64³  | 4.99 s (3 reps) | 5.07 s (3 reps) |
| 256³ | 47.28 s (3 reps) | 46.60 s (3 reps) |

<details>
<summary>Why the GPU FFT doesn't help registration (the key finding)</summary>

- **Output is bit-identical.** Comparing the two `0Warp.nii.gz` displacement fields: `max|Δ| = 0.000` (max warp magnitude 7.1 mm). The FFT backend changes nothing in the result.
- **The GPU sat idle.** `ioreg` Device-Utilization sampling during the 256³ VkFFT run: 3/250 samples nonzero, max 8%. The CPU (metric + gradient + resampling) is the bottleneck; FFT is a sliver of each iteration.
- **SyN field smoothing isn't FFT-based.** `itkGaussianSmoothingOnUpdateDisplacementFieldTransform` smooths the update/total displacement field with `itkGaussianOperator` + separable `itkVectorNeighborhoodOperatorImageFilter` (spatial convolution), **not** `DiscreteGaussianImageFilter`. So the `FFTDiscreteGaussianImageFilterFactory` override only affects the handful of per-level *image* smoothings (→ the few GPU blips), never the per-iteration field smoothing.

Net: to actually benefit from GPU FFT in registration, the FFT would have to sit on a hot path that is FFT-based — the current SyN smoothing is separable spatial convolution, which is already cheap on CPU. Consistent with the standalone numbers where PocketFFT is competitive at ≤256³.

</details>

<details>
<summary>Setup / reproduction</summary>

- Backend selected at runtime via `ANTS_USE_VKFFT` env var (registration helper registers the VkFFT factories at front only when set; FFT-based Gaussian factory always registered). One build, two runs.
- 64³: BRAINSTools `DEBUGFixed/MovingVolume_BF`. 256³: `avg_t1` fixed, fixed rolled by (6,4,-5) voxels as moving.
- `-t SyN[0.25,3.0,0.5] -c [40x20,1e-9,15] -s 1x0vox -f 2x1` (256³).
- Standalone FFT microbench on the same hardware (for reference): VkFFT-Metal 256³ ≈ 8.7 ms vs PocketFFT ≈ 12.5 ms — a real FFT win that simply isn't exercised here.

</details>

hjmjohnson:
Two findings: 

1) PocketFFT  (ITK default now in latest ITK v6) and GPU and FFTW are all nearly the same (GPU computation is faster, but data transfer to/from GPU is about the same as CPU computation time).  

2) RecursiveGaussian on CPU is fastest smoother, more than any FFT variant.

ntustison:
Thank you very much for this @hjmjohnson

cookpa:
Very interesting, thanks. More GPU support to come soon via [FireANTs](https://fireants.readthedocs.io/en/latest/)

PR #1350, opened 2022-05-04
ENH: Allow smoothing to be specified in mm or vox

cookpa:
Previously, affine transforms used the registration method default (mm) while deformable transforms used voxels. 

Made them consistent with antsRegistration, allowing AxBxCmm or AxBxCvox specification.

### DIFF
--- a/Examples/antsMotionCorr.cxx
+++ b/Examples/antsMotionCorr.cxx
@@ -872,8 +872,30 @@ ants_motion(itk::ants::CommandLineParser * parser)
 
     // Get smoothing sigmas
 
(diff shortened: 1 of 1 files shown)


cookpa:
This is for antsMotionCorr, I probably should have clarified in the comment

cookpa:
The default is now voxels, consistent with `antsRegistration`. It was inconsistent before in the code, an affine transform specified smoothing in mm, and SyN smoothing was in voxels. 

A default of mm is more backwards compatible for most use cases, but then we're committing to being different from `antsRegistration`, which is its own problem.

Issue #1348, opened 2022-05-02
Mattes metric fails for very small voxels


cookpa:
Related to ANTsPy issue https://github.com/ANTsX/ANTsPy/issues/362

@ntustison , I found that `antsRegistration` and `antsMotionCorr` will fail with very small voxels. In the attached archive, I used RNifti to edit the pixel spacing of the MNI152NLin6Asym template from 2mm to 0.2mm, 0.02mm, 0.002mm, etc. Once the voxel spacing gets sufficiently small, the registration starts to fail with Mattes, but still works with CC or MSQ.

With mask (I included the mask because antspy's motion correction uses a mask by default)

```
z="02mm" # works with 0.2mm spacing
# z="0002mm" # fails
antsRegistration -d 3 -m Mattes\[ t1w_${z}.nii.gz, t1w_${z}.nii.gz, 1, 32 \] -f 1x1 -s 0x0 -c 1x1 -t Rigid\[ 0.1 \] -x mask_${z}.nii.gz -o reg/result_${z}_ --verbose
```

Without the mask, it goes one level smaller before failing

```
z="0002mm"
# z="00002mm" # fails
antsRegistration -d 3 -m Mattes\[ t1w_${z}.nii.gz, t1w_${z}.nii.gz, 1, 32 \] -f 1x1 -s 0x0 -c 1x1 -t Rigid\[ 0.1 \]  -o reg/result_${z}_ --verbose
```

Data here: [innerSpace.zip](https://github.com/ANTsX/ANTs/files/8605809/innerSpace.zip)

ANTs Version: v2.3.5.post101-g1c5ae7c
Compiled: Mar 23 2022 11:49:51

cookpa:
This is not a perfect test because the images are identical, and the original issue in antspy is intermittent...but I ran it 100 times with the smallest voxels and Mattes fails consistently, and CC does not fail at all.

cookpa:
Reproduced on Russ' data:

```
ImageMath 4 series-tdtomato_size-mm_units-mm_vols/series-tdtomato_size-mm_units-mm_vol_.nii.gz TimeSeriesDisassemble ants_example_data/series-tdtomato_size-mm_units-mm.nii

[...]

PR #1351, opened 2022-05-04
ENH:  Expose gradient filter option for metrics.

ntustison:
Partially attempts to address [this issue](https://github.com/ANTsX/ANTs/issues/1348).  Expose gradient filter usage option for metrics.


gdevenyi:
Any idea how much performance penalty this adds?

PR #1361, opened 2022-05-12
STYLE: Catch exceptions by _const_ reference

hjmjohnson:
Following C++ Core Guidelines, April 10, 2022, which says that it is "typically
better" to catch by const reference than to catch by non-const reference. At
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#e15-throw-by-value-catch-exceptions-from-a-hierarchy-by-reference

Discussed at C++ Core Guidelines issue isocpp/CppCoreGuidelines#518 "Catch
exceptions by const reference?", from February 3, 2016.

find . -type f | egrep '\.[it]?(cc|hh|[ch](\+\+|pp?|xx)?)$' | fgrep -v ThirdParty |     xargs sed -r -i         -e 's/catch \((const )?/catch (const /g'         -e 's/catch \(const \.\.\.\)/catch (...)/g'

### DIFF
--- a/Examples/ANTS.cxx
+++ b/Examples/ANTS.cxx
@@ -87,7 +87,7 @@ ANTSex(int argc, char * argv[])
   {
     registration->RunRegistration();
   }
-  catch (std::exception const & e)
+  catch (const std::exception const & e)
   {
     std::cerr << "Exception caught in ANTS: " << std::endl << e.what() << std::endl;
     return EXIT_FAILURE;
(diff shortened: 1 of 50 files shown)


ntustison:
Awesome.  Thanks @hjmjohnson

cookpa:
Looks like an extra const somewhere, I'll check this out and try to fix the other warning I introduced to antsMotionCorr

Examples/ANTS.cxx
```diff
@@ -87,7 +87,7 @@ ANTSex(int argc, char * argv[])
   {
     registration->RunRegistration();
   }
-  catch (std::exception const & e)
+  catch (const std::exception const & e)
```
cookpa, line 90: This looks like the only error

cookpa:
Sorry @hjmjohnson I might be doing something wrong or you didn't check the "allow edits by maintainers" box on the PR, but I can't seem to push to this. But I did compile so the extra "const" in ANTS.cxx is the only thing preventing the build.

cookpa:
Going to merge and make that one fix

PR #1368, opened 2022-05-24
Const exceptions

cookpa:
Fixing one extra "const" that was preventing the merge of #1361.


cookpa:
Thanks @hjmjohnson for these fixes

PR #1373, opened 2022-06-08
ENH: Denoise all anatomical images, but outside of loop

cookpa:
Before:
 * Only the first image gets denoised
 * Denoising happens at every iteration

Now:
 * Denoise all images
 * Do denoising once outside of the loop

Also added a check for prior images - if priors are incorrectly specified, exit with error instead of carrying on with KMeans.


ntustison:
Thanks @cookpa

PR #1389, opened 2022-07-11
ENH: Progress reporting for antsAI

### LINKED ISSUE #439: Feature Request: more verbosity for antsAI
antsAI's --verbose option isn't particularly verbose, particularly in the area of showing progress.

It's tough to gauge when a call might be done with no output.

cookpa:
Proof of concept for progress reporting in `antsAI`. Currently prints every iteration, which is potentially a lot - could be smarter and print progress at intervals.

Fixes #439


cookpa:
Updated with some improvements. Outputs to terminal for every [(total iterations) / 10] iterations, and reports the best metric value at the end.

akvilonBrown:
Thank you!
 https://github.com/ANTsX/ANTs/issues/1529

PR #1393, opened 2022-07-14
ENH: Use docker to build Centos binaries

cookpa:
This can be extended to other platforms that aren't supported by native runners.

Also made some improvements for readability to the native release binary build

Swapped out the artifact uploader to one that retries if there's a network error

### DIFF
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -16,44 +16,64 @@ jobs:
       matrix:
         config:
         - {
-            name: "Ubuntu_20.04-GCC",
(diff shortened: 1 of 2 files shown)


cookpa:
To do: Include architecture in binary names.

cookpa:
I made a bunch of changes to do things better, added more targets so we pretty much build on all supported Mac and Linux runners now. 

Linux runners are all Ubuntu, but using Docker, I was able to build inside Centos 7. We can add other targets the same way, as long as they build in time. The Centos build is slightly slower because the runner crashes out of memory with a parallel build, but it's well under the six hour limit.

PR #1406, opened 2022-07-28
ENH: Allow fixed,moving masks in scripts

cookpa:
This was kind of an undocumented feature because you could do `-x fixedMask.nii.gz,movingMask.nii.gz` and it would work. The script would change the arg to `-x [ fixedMask.nii.gz,movingMask.nii.gz, NULL ]`. 

I changed it so that `-x fixedMask.nii.gz,movingMask.nii.gz` maps to `-x [ fixedMask.nii.gz,movingMask.nii.gz ]` and `-x fixedMask.nii.gz` maps to `-x [ fixedMask.nii.gz, NULL ]` as before.

### DIFF
--- a/Scripts/antsRegistrationSyN.sh
+++ b/Scripts/antsRegistrationSyN.sh
@@ -82,10 +82,10 @@ Optional arguments:
 
      -g:  gradient step size for SyN and B-spline SyN (default = 0.1)
 
-     -x:  mask(s) for the fixed image space.  Should specify either a single image to be used for
-          all stages or one should specify a mask image for each "stage" (cf -t option).  If
-          no mask is to be used for a particular stage, the keyword 'NULL' should be used
-          in place of a file name.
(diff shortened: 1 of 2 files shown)


cookpa:
Not easily becasue NULL isn't just a script convention, it has special meaning in the C++ code.

gdevenyi:
The way the code works in the masking inside antsRegistration seems to be 
- check if file exists and is loadable as a mask
- if not loadable use no mask

Ergo, you can create a file called NULL and antsRegistration will use it (ignoring the complications of file format detection...)

cookpa:
Oh yeah, so it does. It gets the nullptr from attempting to read the file. 

I guess then it could be changed. But now I'm thinking it would be better to parse NULL without printing the error message - because the code as written will run without a mask if there is a typo in the mask file name.

gdevenyi:
This would cause me much pain because I've been using NOMASK for a wide swath of scripts I'd need to update.

I stil support making it error out if it can't read the mask provided.

Issue #1405, opened 2022-07-26
No impact of the --use-estimate-learning-rate-once (-l) flag


devDonnn:
The flag `--use-estimate-learning-rate-once` does not seems to provide any changes to the final output of the `antsRegistration`. Isn't it fully implemented? I see it is part of the `parser` but cannot find where it is implemented. Thanks!

cookpa:
https://github.com/ANTsX/ANTs/blob/04a018dc5308183b455194a9a5b14ffe1b0edf5f/Examples/antsRegistrationTemplateHeader.h#L442-L454

devDonnn:
Thanks. I have found this too but why this option does not have any impact on the final results?

cookpa:
It doesn't have any impact at all? You get the exact same results both ways?

devDonnn:
Yes. I ran twice in repro mode with `--use-estimate-learning-rate-once` and without. The verbose output did not change as well.

cookpa:
That's interesting. Can you share a reproducible example with code and data?

devDonnn:
I have used the following code to enable `--use-estimate-learning-rate-once`:


antsRegistration 
  --verbose 1 
  --dimensionality 3 
  --float 0 
  --collapse-output-transforms 1 
  --output [ movingToFixed_, movingToFixed_Warped.nii.gz, movingToFixed_ InverseWarped.nii.gz ] 
  --interpolation Linear 
  --use-histogram-matching 0 
  --use-estimate-learning-rate-once 1 
  --winsorize-image-intensities [ 0.005,0.995 ] 
  
  --initial-moving-transform [ fixedImage.nii.gz,movingImage.nii.gz,1 ] 
  
  --transform Rigid[ 0.1 ] 
  --metric MI[ fixedImage.nii.gz,movingImage.nii.gz,1,32,Regular,0.25 ] 
  --convergence [ 1000x500x250x100,1e-6,10 ] 
  --shrink-factors 8x4x2x1 
  --smoothing-sigmas 3x2x1x0vox 
 
  --transform Affine[ 0.1 ] 
  --metric MI[ fixedImage.nii.gz,movingImage.nii.gz,1,32,Regular,0.25 ] 
  --convergence [ 1000x500x250x100,1e-6,10 ] 
  --shrink-factors 8x4x2x1 
  --smoothing-sigmas 3x2x1x0vox 
  
  --transform SyN[ 0.1,3,0 ] 
[...]

PR #1411, opened 2022-08-09
WIP:  Remove unused estimate learning rate at each iteration.

ntustison:
# Description

I'm guessing this is a legacy option before we switched to the conjugate gradient descent optimizers for everything but the exponential transforms.   But even those don't use this flag as no scales estimator is applied in these two cases (B-spline and Gaussian).

Addresses #1405

### DIFF
--- a/Examples/antsRegistration.cxx
+++ b/Examples/antsRegistration.cxx
@@ -193,7 +193,6 @@ antsRegistrationInitializeCommandLineOptions(itk::ants::CommandLineParser * pars
     option->SetShortName('z');
     option->SetUsageOption(0, "(1)/0");
     option->SetDescription(description);
-    option->AddFunction(std::string("1"));
     parser->AddOption(option);
   }
 
@@ -445,17 +444,16 @@ antsRegistrationInitializeCommandLineOptions(itk::ants::CommandLineParser * pars
     parser->AddOption(option);
   }
 
-  {
-    std::string description =
-      std::string("turn on the option that lets you estimate the learning rate step size only at the beginning of each "
-                  "level.  * useful as a second stage of fine-scale registration.");
-
-    OptionType::Pointer option = OptionType::New();
-    option->SetLongName("use-estimate-learning-rate-once");
-    option->SetShortName('l');
-    option->SetDescription(description);
-    parser->AddOption(option);
-  }
+  // {
+  //   std::string description =
+  //     std::string("Turn on the option that lets you estimate the learning rate step size only at the beginning of each "
+  //                 "level.  This is useful as a second stage of fine-scale registration.");
+  //   OptionType::Pointer option = OptionType::New();
+  //   option->SetLongName("use-estimate-learning-rate-once");
+  //   option->SetShortName('l');
+  //   option->SetDescription(description);
+  //   parser->AddOption(option);
+  // }
 
   {
     std::string description = std::string("Winsorize data based on specified quantiles.");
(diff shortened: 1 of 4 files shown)


cookpa:
Thanks @ntustison ! 

I think this means we should also remove this option from antsMotionCorr.cxx, which also uses a conjugate gradient optimizer

https://github.com/ANTsX/ANTs/blob/master/Examples/antsMotionCorr.cxx#L131

Examples/antsRegistration.cxx
```diff
@@ -193,7 +193,6 @@ antsRegistrationInitializeCommandLineOptions(itk::ants::CommandLineParser * pars
     option->SetShortName('z');
     option->SetUsageOption(0, "(1)/0");
     option->SetDescription(description);
-    option->AddFunction(std::string("1"));
```
cookpa, line 196: I don't think this should have been removed, I get a seg fault now unless I specify `--collapse-output-transforms 1`. The argument handling is inconsistent and probably needs to be refactored, but some args (like this) have the convention that a default is assumed. In others, there is a check for the number of functions and only if > 0 is the value queried.

ntustison:
Thanks for trying this.  Can you tell me exactly the command you're running because I can't reproduce?

Examples/antsRegistration.cxx
stnava: I this going to impact ANTsR and ANTsPy?  I believe these do set the option internally.

cookpa:
bash -c "../install/bin/antsRegistration --verbose 1 --dimensionality 3 --float 0 --output [ test,testWarped.nii.gz,testInverseWarped.nii.gz ] --interpolation Linear --use-histogram-matching 0 --winsorize-image-intensities [ 0.005,0.995 ] --initial-moving-transform [ tpl-MNI152NLin2009cAsym_res-01_T1w.nii.gz,T1w.nii.gz,1 ] --transform Rigid[ 0.1 ] --metric MI[ tpl-MNI152NLin2009cAsym_res-01_T1w.nii.gz,T1w.nii.gz,1,32,Regular,0.25 ] --convergence [ 100x50x25x10,1e-6,10 ] --shrink-factors 8x4x2x1 --smoothing-sigmas 3x2x1x0vox --transform Affine[ 0.1 ] --metric MI[ tpl-MNI152NLin2009cAsym_res-01_T1w.nii.gz,T1w.nii.gz,1,32,Regular,0.25 ] --convergence [ 100x50x25x10,1e-6,10 ] --shrink-factors 8x4x2x1 --smoothing-sigmas 3x2x1x0vox --transform SyN[ 0.1,3,0 ] --metric CC[ tpl-MNI152NLin2009cAsym_res-01_T1w.nii.gz,T1w.nii.gz,1,4 ] --convergence [ 10x7x5x0,1e-6,10 ] --shrink-factors 8x4x2x1 --smoothing-sigmas 3x2x1x0vox"

This doesn't work, but it does work if I add `--collapse-output-transforms 1`

ntustison:
Thanks @cookpa , I'll look at it in a few minutes.

An alternative is to just add documentation to the option.  I'm fine with whatever.

cookpa (commented):
This works for me

ntustison:
Thanks @cookpa , @stnava .  I'll take a look at any ANTsPy/ANTsR dependencies later today before merging.

ntustison:
Also, I'll take a look at the other programs (eg, antsMotionCorr).

cookpa:
Thanks for sorting this @ntustison

cookpa:
I think it was 2013, here

https://github.com/ANTsX/ANTs/commit/42dd9d3f1f86778327de9aeb6789ce4ddd34d1fe

PR #1415, opened 2022-08-23
ENH: Output extra info at the end of registration

cookpa:
It's easy to miss the printout of the antsRegistration call when using the scripts, because it gets scrolled off the screen. So I repeated it at the end.


ntustison:
Great idea.

cookpa:
Thanks @ntustison. I am thinking to do a patch release tonight. If all goes well we will have compiled binaries attached automatically.

PR #1419, opened 2022-09-06
DOC: added missing end parenthesis to Atropos documentation string

br-cpvc:
added missing end parenthesis to Atropos' use-euclidean-distance parameter documentation.


br-cpvc:
accidentally delete the branch because I renamed it, made a new pr as https://github.com/ANTsX/ANTs/pull/1421 with the same change

Issue #1017, opened 2020-06-04
Limited Use of Multithreading during SyN Registration (b-spline syn)


NitramNimod:
Hi!
When using b-spline syn during ANTs registration only one cpu core is working, while normal SyN behaves in the way it should behave, when properly setting and using the ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS variable. I can see in the resource monitor a short cpu work load spike right at the beginning of each convergence step, but 99% of the time it's single core processing.
This behavior is reproducible on a Mac running El Capitan (10.11.6) or an Ubuntu 18.04.4 LTS (GNU/Linux 5.3.0-53-generic x86_64).
The Mac is running the latest ANTs version (2.3.3.dev168-g29bdf) compiled this morning, on Ubuntu slightly older (2.3.1.dev120-g2f09e) compiled in March.
I've found an older similar thread which is already closed because the problem was solved by compiling against a newer ITK, which here obviously doesn't work. [https://github.com/ANTsX/ANTs/issues/604]
As the b-spline syn takes a lot more time a multi-threading would be highly appreciated! :)

Thanks and best regards,
Martin

ntustison:
All multi-threading is handled by ITK.  The SyN and B-spline SyN classes are derived from the same parent class with the only difference being the use of B-spline smoothing and that B-spline smoothing class has been there for ~10 years. Did you try compiling a fresh checkout directly on your machine?

NitramNimod:
Yes, I was using the small installation script that is offered on the ANTs website. I suppose, the git command there will checkout the latest working version. For the Mac a few hours ago…
There were no errors during compilation, everything else is working normal.
How to solve this? Or what can I do to help?

ntustison:
It's really hard to say what's going on because, as I said, multi-threading is handled by the ITK foundation.  You might want to ask over at the ITK discussion forum.

cookpa:
I've not come across similar problems since that old thread. Happy to test again with a reproducible example. But my knowledge of the underlying ITK behavior is limited, so we'd probably have to ask over there.

NitramNimod:
Example in the sense of "data" or "parameters" for an antsRegistration-call?

NitramNimod:
Ok, I've assembled a simple example, I just used the famous "anatomy of an antsRegistration call" call and changed the SyN stage to BSplineSyN:

antsRegistration --dimensionality 3 --float 0 --verbose 1 \\
[...]

PR #1420, opened 2022-09-06
ENH: antsAtroposN4.sh - exposed more subcommand parameters as arguments

br-cpvc:
when trying to optimize antsAtroposN4.sh execution speed contra output precision it would be great to expose these internal variables as external cli parameters

### DIFF
--- a/Scripts/antsAtroposN4.sh
+++ b/Scripts/antsAtroposN4.sh
@@ -78,6 +78,7 @@ Optional arguments:
 
      -e: N4 convergence, e.g. "[25x25x25x25,0]"
      -f: N4 shrink factor
+     -q: N4 B-spline parameters
      -i: Atropos icm useAsynchronousUpdate 0/(1)
      -j: Atropos use-euclidean-distance (0)/1
 
@@ -208,7 +209,7 @@ if [[ $# -lt 3 ]] ; then
   Usage >&2
   exit 1
 else
-  while getopts "a:b:c:d:e:f:g:h:i:j:k:l:m:n:o:p:r:s:t:u:w:x:y:z:" OPT
+  while getopts "a:b:c:d:e:f:g:h:i:j:k:l:m:n:o:p:q:r:s:t:u:w:x:y:z:" OPT
     do
       case $OPT in
           c) #number of segmentation classes
@@ -237,6 +238,9 @@ else
               ;;
           f)
               N4_SHRINK_FACTOR=$OPTARG
+              ;;
+          q)
+              N4_BSPLINE_PARAMS=$OPTARG
               ;;
           g) # denoise anatomical images
        DENOISE_ANATOMICAL_IMAGES=$OPTARG
--- a/Scripts/antsAtroposN4.sh
+++ b/Scripts/antsAtroposN4.sh
@@ -76,6 +76,11 @@ Optional arguments:
      -u:  use random seeding                    Use random number generated from system clock in Atropos (default = 1)
      -w:  Atropos prior segmentation weight     Atropos spatial prior probability weight for the segmentation (default = 0)
 
+     -e: N4 convergence, e.g. "[25x25x25x25,0]"
+     -f: N4 shrink factor
+     -i: Atropos icm useAsynchronousUpdate 0/(1)
+     -j: Atropos use-euclidean-distance (0)/1
+
      -z:  Test / debug mode                     If > 0, attempts to continue after errors.
 
 USAGE
@@ -108,6 +113,8 @@ echoParameters() {
        Max N4->Atropos iters. = ${N4_ATROPOS_NUMBER_OF_ITERATIONS}
        Max Atropos iters.     = ${ATROPOS_SEGMENTATION_NUMBER_OF_ITERATIONS}
        use clock random seed  = ${USE_RANDOM_SEEDING}
+       icm                    = ${ATROPOS_SEGMENTATION_ICM}
+       use-euclidean-distance = ${ATROPOS_SEGMENTATION_USE_EUCLIDEAN_DISTANCE}
 
 PARAMETERS
 }
@@ -192,6 +199,8 @@ ATROPOS_SEGMENTATION_POSTERIOR_FORMULATION="Socrates[ 1 ]"
 ATROPOS_SEGMENTATION_MASK=''
 ATROPOS_SEGMENTATION_NUMBER_OF_ITERATIONS=5
 ATROPOS_SEGMENTATION_NUMBER_OF_CLASSES=3
+ATROPOS_SEGMENTATION_ICM=1
+ATROPOS_SEGMENTATION_USE_EUCLIDEAN_DISTANCE=0
 ATROPOS_SEGMENTATION_MRF=''
 ATROPOS_SEGMENTATION_LABEL_PROPAGATION=()
 
@@ -199,7 +208,7 @@ if [[ $# -lt 3 ]] ; then
   Usage >&2
   exit 1
 else
-  while getopts "a:b:c:d:g:h:k:l:m:n:o:p:r:s:t:u:w:x:y:z:" OPT
+  while getopts "a:b:c:d:e:f:g:h:i:j:k:l:m:n:o:p:r:s:t:u:w:x:y:z:" OPT
     do
       case $OPT in
           c) #number of segmentation classes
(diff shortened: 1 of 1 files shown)


cookpa:
Thanks for this.

To do:

- [ ] Verify script is compatible with existing antsCorticalThickness.sh call (same options get used)
- [ ] Verify non-default options get passed through correctly (eg, Atropos -g 1 works and doesn't have to be -g [ 1 ])

gdevenyi:
> when trying to optimize antsAtroposN4.sh execution speed contra output precision

Do you have any docs/experiments on why these Atropos options speed things up/what they do? Atropos is a bit of black box as it has a ton of things not mentioned in the paper which introduces it.

br-cpvc:
@gdevenyi That is a very good question :) I would also very much appreciate more information/documentation on what the different Atropos parameters actually affect. Unfortunately I do not have the enough insights or knowledge to fully explain it, but I can provide what insight I have gained during my work.

We have been testing out different combinations of parameters and have currently found that the following parameters yield the best trade off between execution speed and output quality for our input images: Atropos parameters "-e 1 -g 0" (for the N4BiasFieldCorrection we use the parameters "-s 4 -c [35x35x35x35,0] -b [200]").

In regards to why the "-g 0" Atropos paramter accelerate the execution speed: This sets the m_UseAsynchronousUpdating=false in the antsAtroposSegmentationImageFilter.hxx filter used by Atropos, which skips the computations inside the if statement in:
https://github.com/ANTsX/ANTs/blob/77b8f2ad1fa0c4172edba451e8e52d3a1bff33ac/ImageSegmentation/antsAtroposSegmentationImageFilter.hxx#L1153-L1206

In relation to the "-e 1" parameter. As I understand the "Atropos --help" text:

https://github.com/ANTsX/ANTs/blob/77b8f2ad1fa0c4172edba451e8e52d3a1bff33ac/Examples/Atropos.cxx#L1536-L1544
and from the code in:

https://github.com/ANTsX/ANTs/blob/77b8f2ad1fa0c4172edba451e8e52d3a1bff33ac/ImageSegmentation/antsAtroposSegmentationImageFilter.h#L611-L617

the best results are found when "-e 1", which is in line what we have observed.

The exact difference in computation between the two can be seen here:
https://github.com/ANTsX/ANTs/blob/77b8f2ad1fa0c4172edba451e8e52d3a1bff33ac/ImageSegmentation/antsAtroposSegmentationImageFilter.hxx#L2433-L2513

gdevenyi:
Thanks @br-cpvc as I suspected reading the code is still the only way to determine some of the effects here. The comment about the distance calculation is particularly helpful. Cheers.

br-cpvc:
@cookpa I initially thought that the "To do list" that you inserted was for your self or someone on your team, but as no updates have been posted I have started to wonder if it was ment for me? Are you waiting for me to do something or what is the current status?

cookpa:
It's for anyone who has time to do it, you're welcome to take it on if you'd like. 

Since you have exposed some N4 parameters, it might be nice to add the initial mesh resolution / spacing (currently hard coded as `N4_BSPLINE_PARAMS`). That would make the script more applicable to smaller brains.

BTW, if you're benchmarking performance, there have been recent changes in ITK that might affect multi-threading performance in N4 (#1017).

br-cpvc:
@cookpa thank you for the comment, and thank you for making me aware of the recent patch (#1017) affect multi-threading performance in N4.

I have looked into what it takes to align the input parameters between the two scripts.

With the current getopts argument parser it seems impossible to add and/or align the input parameters between antsAtroposN4.sh and antsCorticalThickness.sh - this is because of two things:
- getopts only supports single-letter options parsing
- antsCorticalThickness.sh have already used up all single-letter option letters (except -h)

I think the best will be to make it possible to use long options, but as getopts does not support this, another approach to parsing the arguments must be used.

I first looked at the getopt parameter parser, which supports long options. But of what I could find, the getopt package is not recommended, see: http://mywiki.wooledge.org/ComplexOptionParsing the only other option I could find was to implement the parsing directly/manually inside the script, which I have tried based on the example: https://mywiki.wooledge.org/BashFAQ/035#Manual_loop

I have redone my implementation based on the current head of the ANTs main branch, and included a new cli option for setting the N4_BSPLINE_PARAMS as you suggested. The reimplementation is currently in https://github.com/ANTsX/ANTs/compare/master...br-cpvc:ANTs:upstream_20221110_0945

What do you think of the approach? Is this in the right direction or should it be done another way?

gdevenyi:
The solution for complex bash parsing is https://argbash.readthedocs.io/en/latest/ and its fantastic.

cookpa:
Yeah, I'm not sure it's feasible to make every option used within `antsCorticalThickness.sh` available, without completely rewriting the option parsing. I really just want to verify that changes to `antsAtroposN4.sh` don't result in changes how that script behaves when called from within `antsCorticalThickness.sh`

br-cpvc:
@cookpa Ah, okay, then I completely misunderstood what you were after :)

I have now taken another go at it.

I have pushed another patch to the current PR branch which adds the N4_BSPLINE_PARAMS parameters as you previously suggested. After that, I made a test trying to show that any script that uses antsAtroposN4.sh should expect the same results as before the changes.

The test shows that antsAtroposN4.sh produces the same results (it does not alter the output). By executing it two times, before and after the change to the Atropos parameters and their default values (ATROPOS_SEGMENTATION_ICM=1 and ATROPOS_SEGMENTATION_USE_EUCLIDEAN_DISTANCE=0) which were added to the exe_segmentation statement in antsAtroposN4.sh https://github.com/ANTsX/ANTs/compare/master...br-cpvc:ANTs:upstream_20220906_1020#diff-b35ac52557562254e75ede77f4c181598f938e248bcfcd8d7c4836c7a7d1048dL502-L503

The test:
- execute antsAtroposN4.sh with -u 0 using git checkout before the changes (e043d8a690be769cff7656f3747bb803cca7d24a)
- execute antsAtroposN4.sh with -u 0 using git checkout after the changes (fcb7aa60c9dd036f884e3d6f28e2c0f6fef488d0)
- compare the output using md5sum


Execution of the test
```

git checkout e043d8a690be769cff7656f3747bb803cca7d24a
mkdir -p processing_output/baseline
./Scripts/antsAtroposN4.sh -d 3 -u 0 -a sub-0303:anat:sub-0303_T1w.nii.gz -x sub-0303:anat:sub-0303_T1w.nii.gz -o processing_output/baseline/


git checkout fcb7aa60c9dd036f884e3d6f28e2c0f6fef488d0
mkdir -p processing_output/upstream_20220906_1020
./Scripts/antsAtroposN4.sh -d 3 -u 0 -a sub-0303:anat:sub-0303_T1w.nii.gz -x sub-0303:anat:sub-0303_T1w.nii.gz -o processing_output/upstream_20220906_1020/

md5sum processing_output/*/*
4404b487734935d2bb5a455e5f772e7b  processing_output/baseline/Segmentation.nii.gz
b2f3be2cf952cb124edc94096d64991e  processing_output/baseline/Segmentation0N4.nii.gz
dc4780156a987295ee98ba66afb30b59  processing_output/baseline/SegmentationConvergence.txt
67772085703eba7367614918f4548792  processing_output/baseline/SegmentationPosteriors1.nii.gz
1385edf67ea6c11d05d061bfea1c4476  processing_output/baseline/SegmentationPosteriors2.nii.gz
472f5cc135ed93b0fae8a615966086d9  processing_output/baseline/SegmentationPosteriors3.nii.gz

4404b487734935d2bb5a455e5f772e7b  processing_output/upstream_20220906_1020/Segmentation.nii.gz
b2f3be2cf952cb124edc94096d64991e  processing_output/upstream_20220906_1020/Segmentation0N4.nii.gz
dc4780156a987295ee98ba66afb30b59  processing_output/upstream_20220906_1020/SegmentationConvergence.txt
67772085703eba7367614918f4548792  processing_output/upstream_20220906_1020/SegmentationPosteriors1.nii.gz
1385edf67ea6c11d05d061bfea1c4476  processing_output/upstream_20220906_1020/SegmentationPosteriors2.nii.gz
472f5cc135ed93b0fae8a615966086d9  processing_output/upstream_20220906_1020/SegmentationPosteriors3.nii.gz
```
Both outputs are exacly the same.

I also tried changing the ICM parameter to see that it actually change the results:
```
./Scripts/antsAtroposN4.sh -d 3 -u 0 -a sub-0303:anat:sub-0303_T1w.nii.gz -x sub-0303:anat:sub-0303_T1w.nii.gz -o processing_output/upstream_20220906_1020_disabled/ -i 0

md5sum processing_output/upstream_20220906_1020_disabled/*
215573439fa491dd3b49bd9617d0494f  processing_output/upstream_20220906_1020_disabled/Segmentation.nii.gz
b2f3be2cf952cb124edc94096d64991e  processing_output/upstream_20220906_1020_disabled/Segmentation0N4.nii.gz
8b8cd3587688ea93b81b30f45151b7b8  processing_output/upstream_20220906_1020_disabled/SegmentationConvergence.txt
d9e21f674b42a3d4a3014da272fbe70d  processing_output/upstream_20220906_1020_disabled/SegmentationPosteriors1.nii.gz
be2190411d00843e6d78c54e7e5f7144  processing_output/upstream_20220906_1020_disabled/SegmentationPosteriors2.nii.gz
503d74d7485987bc634ff394d421e9d5  processing_output/upstream_20220906_1020_disabled/SegmentationPosteriors3.nii.gz
```

What do you think, is this test deep enough to cover what you were asking?

cookpa:
Yes that looks good to me, thanks. I did a similar test and got identical results with the new script. So I think this is suitable to merge.

PR #1422, opened 2022-09-06
ENH: graceful exit Atropos in case of non-existing input files

br-cpvc:


cookpa (commented):
Would you mind also getting the ReadImage call on L267

https://github.com/ANTsX/ANTs/blob/04438e5d4bb44c4b85efd5e8ad2f411c17151f0b/Examples/Atropos.cxx#L267

br-cpvc:
@cookpa thanks for review the code, and your response. I have added a fixup patch to the branch, this patch includes the sugested change to L267. The patch also include a similar change in L221 where this was also a problem.

cookpa:
LGTM thank you

PR #1436, opened 2022-10-21
ENH: Check AverageImages input files can be read

cookpa:
DOC: Clarify normalization options only apply to single component input


cookpa:
This replaces a segfault with an error condition and meaningful error message.

PR #1445, opened 2022-11-04
COMP: Pass optimization flags to ITK and ANTS-build

cookpa:
SuperBuild_ANTS_C[XX]_OPTIMIZATION_FLAGS were defined as options, but were not being used. The project optimization flags were reset to their defaults by the inclusion of ITKSetStandardCompilerFlags.cmake. Now pass these values to the subprojects so consistent optimization flags are used for ITK and ANTs. I could not find an equivalent setting for VTK.


ntustison:
Thanks @cookpa

PR #1452, opened 2022-11-21
PERF: openmp accelerating antsAtroposSegmentationImageFilter.hxx function GetPosteriorProbabilityImage

br-cpvc:
These code changes shows the idear of an optimization that uses openmp to accelerate the antsAtroposSegmentationImageFilter.hxx function GetPosteriorProbabilityImage. I have tried to implement a patch with as minimal changes to the original code as possible. Currently the code only works for the 3D case and reports errors for the other dimensionalities.

During my testing I have found that the optimization reduces the Atropos execution time in the range 5% to 38% depending on the input.

My first question is, if this is the right approach for including parallel for-loop acceleration in ANTs?, and further if this approach is thought of as good practice for the ANTs code base or if another approach should be used? Also, are optimizations like these even likely to be accepted into the ANTs code base? If it is, I would be glad to rework the code into an acceptable state if possible.

I'm however currently unsure how best to generalise this optimization without having to implement a special case for each dimension (this optimization sits in the middle/between c++ code structures and it is not easy to implement dymanically/at runtime switching between the optimized and not optimized code pathes based on the ImageDimension (without copy/pasting a lot of code). I could really use some suggestions on how to generalise the work for all supported dimensionlities (2,3,4) without too many special cases. If possible my goal is to add this acceleration with minimum impact on the current code base.

### DIFF
--- a/ImageSegmentation/antsAtroposSegmentationImageFilter.hxx
+++ b/ImageSegmentation/antsAtroposSegmentationImageFilter.hxx
@@ -1988,11 +1988,43 @@ AtroposSegmentationImageFilter<TInputImage, TMaskImage, TClassifiedImage>::GetPo
           radius[d] = this->m_MRFRadius[d];
         }
 
+#define __optimize_omp_GetPosteriorProbabilityImage__
+#ifdef __optimize_omp_GetPosteriorProbabilityImage__
+    if (ImageDimension!=3) {
+      itkWarningMacro( "Only 3 Dimensions are currently supported." );
+      exit(EXIT_FAILURE);
+    }
+std::cout << "optimize_omp_GetPosteriorProbabilityImage" << std::endl;
+
+    unsigned long xSize = this->GetOutput()->GetRequestedRegion().GetSize(0);
+    unsigned long ySize = this->GetOutput()->GetRequestedRegion().GetSize(1);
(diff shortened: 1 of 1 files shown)


br-cpvc:
Hi @gdevenyi Thank you for the pointer to the ITK documentation. I'm trying to wrap my head around the ITK multi-threading framework, but cannot seem to find a way to convert my current openmp accelerated code into code that utilizes to ITK building blocks described in the documentation.

If any one have further ideas, suggestion, or resources such as examples or tutorials, please let me know, any help will be of greatly value.

gdevenyi:
Hi,

I'm just a dabbler in the ITK code, so I don't have sufficient skill to help you directly. I suggest bringing this up at the ITK discourse where the developers are very active, https://discourse.itk.org/

br-cpvc:
@gdevenyi Thank you for you suggestion, I will try doing that :)

br-cpvc:
Created the post: https://discourse.itk.org/t/how-to-use-itk-multi-threading-primitives-to-do-parallel-execution-of-loop-iteration/5531

br-cpvc:
Superseded by reworked code in new PR https://github.com/ANTsX/ANTs/pull/1462

PR #1454, opened 2022-11-21
BUG: Atropos no longer segfaults when -i is missing, it reports an error msg instead

br-cpvc:
bug can be reproduced by:

```
mkdir tmp_test
Atropos -x sub-0303:anat:sub-0303_T1w.nii.gz -a sub-0303:anat:sub-0303_T1w.nii.gz -o [ tmp_test/Segmentation.nii.gz,tmp_test/SegmentationPosteriors%d.nii.gz ]
```

this currently segfaults, the patched code instead prints an error message to explain to the user what is wrong.


cookpa:
Thanks for this @br-cpvc

PR #1468, opened 2022-12-22
ENH: Make "-u" option also set a random seed for registration

cookpa:
This enables identical results from multiple runs of brain extraction by using "-u 0". Previously, this was achieved by setting ANTS_RANDOM_SEED along with "-u 0". Now, "-u 0" also sets ANTS_RANDOM_SEED within the script if it is not already defined


ntustison:
Hey, thanks for this @cookpa

PR #1526, opened 2023-04-20
Enable MINC support by default

gdevenyi:
This is a minimal dependency change for the build (just libminc2 in ITK) and the ANTs tools show `--minc` regardless of this switch so I think its better to enable by default.


gdevenyi:
This will also make the auto-built binaries fully featured.

Issue #1540, opened 2023-05-21
Template dependence on image spacing


ntustison:
Ran into a recent issue where a collaborator arbitrarily multiplied the image spacing by 1000 and generated a significantly different template---much better, in fact.  Data were mice MRI with voxel spacing on the order of 0.05mm^3.  Investigating this a bit, the ``SharpenImage`` step has a dependency on the image spacing (specifically, ``m_UseImageSpacing`` is true by default in the ``itkLaplacianSharpeningImageFilter``).  I'm posting this since a possible fix isn't obvious to me.  Hoping for some thoughts.

cookpa:
I think this is an issue in multiple contexts and maybe in ITK generally as mentioned here for example

https://discourse.itk.org/t/register-two-binary-images-of-different-size-and-resolution/5936/2

Also

https://github.com/ANTsX/ANTsPy/issues/362

That issue was fixed in ANTs by the gradient filters, which I need to port to ANTsPy to close out the issue.

ntustison:
Yes, it's definitely in other places but I think we can specifically address it in the template building.  I'm here at the University of Minnesota with @stnava.  We'll discuss this issue and I'll post what some of our thoughts are.

cookpa:
Some other places where spacing may come into play:

* Default `BSplineSyN` params are `[ 0.1,26,0,3 ]`, would need adjusting for small animals

* N4 uses old convention of 200mm initial spline spacing

ntustison:
Thanks @cookpa. 

Yeah, they don't concern me as much.  Importantly, we're aware of both of these and we can provide guidance to the user.  Specifically, for the former, I think you've mentioned your preference for N4 preprocessing prior to template building (which I would have chosen had I written the original buildtemplate.sh script, i.e., completely remove N4 from template building) and, for the latter, we can always suggest the user employ regular SyN to avoid the image spacing dependence (which is the default anyway).  

When I believed the culprit was the Laplacian sharpening, my motivation for turning off image spacing dependence was the fact that most of our template building experience was based on the ~1mm data---basically equivalent to voxel space operation.  Now I have to dig a bit further to figuring out what's causing the issue in addition to fixing the Laplacian filter on the ITK side.  I'm going to keep this pull request on hold until I investigate further.

spinicist:
[...]

PR #1543, opened 2023-05-23
ENH:  Add spacing option for laplacian sharpening.

ntustison:
Add option to sharpen without using the image spacing.  Explicitly turn off in template building.  Addresses https://github.com/ANTsX/ANTs/issues/1540.


ntustison:
I'm guessing this is a [false positive](https://github.com/InsightSoftwareConsortium/ITK/issues/4051), unfortunately.  ``m_UseImageSpacing`` seems to not even be used in the itkLaplacianSharpeningImageFilter.  Pausing on this until I figure out more.

gdevenyi:
Weren't there some ITK changes done as well requiring a bump in the ITK commit for the superbuild?

ntustison:
Yes, I'm sure we'll bring in those changes at some point, and you're welcome to do so, but this pull commit doesn't change any current behavior.

Issue #1561, opened 2023-07-03
antsRegistration output with initial transforms and no iterations differs from initial transforms


cookpa:
This has been puzzling me. I wanted to test that `--initial-moving-transform` would work with more than one transform. To test this, I tried passing the results of a registration with `--initial-moving-transform` then immediately writing output.

1.  Do a registration

```
antsRegistration --verbose 1 --dimensionality 3 --float 0 --collapse-output-transforms 1 --output [ oasisToMNIQuick,oasisToMNIQuickWarped.nii.gz,oasisToMNIQuickInverseWarped.nii.gz ] --interpolation Linear --use-histogram-matching 0 --winsorize-image-intensities [ 0.005,0.995 ] --initial-moving-transform [ tpl-MNI152NLin2009cAsym_res-01_T1w.nii.gz,tpl-OASIS30ANTs_res-01_T1w.nii.gz,1 ] --transform Rigid[ 0.1 ] --metric MI[ tpl-MNI152NLin2009cAsym_res-01_T1w.nii.gz,tpl-OASIS30ANTs_res-01_T1w.nii.gz,1,32,Regular,0.25 ] --convergence [ 1000x500x250x0,1e-6,10 ] --shrink-factors 8x4x2x1 --smoothing-sigmas 3x2x1x0vox --transform Affine[ 0.1 ] --metric MI[ tpl-MNI152NLin2009cAsym_res-01_T1w.nii.gz,tpl-OASIS30ANTs_res-01_T1w.nii.gz,1,32,Regular,0.25 ] --convergence [ 1000x500x250x0,1e-6,10 ] --shrink-factors 8x4x2x1 --smoothing-sigmas 3x2x1x0vox --transform SyN[ 0.1,3,0 ] --metric MI[ tpl-MNI152NLin2009cAsym_res-01_T1w.nii.gz,tpl-OASIS30ANTs_res-01_T1w.nii.gz,1,32] --convergence [ 100x70x50x10,1e-6,10 ] --shrink-factors 8x4x2x1 --smoothing-sigmas 3x2x1x0vox
```

2. Apply the warps the usual way 

```
antsApplyTransforms -d 3 -i tpl-OASIS30ANTs_res-01_T1w.nii.gz -r tpl-MNI152NLin2009cAsym_res-01_T1w.nii.gz -t oasisToMNIQuick1Warp.nii.gz -t oasisToMNIQuick0GenericAffine.mat -o oasisToMNIQuickAAT.nii.gz
[...]

PR #1567, opened 2023-07-04
BUG: Avoid composite transform mangling.

ntustison:
https://github.com/ANTsX/ANTs/issues/1561#issuecomment-1619077395

Try this.  

When "collapsing the transforms" the composite transform member variable is getting mangled somehow.  The output transforms written to disk appear to be correct but the mangling affects the warped image so I simply reordered the output which admittedly is a temporary fix but a potential fix nonetheless.


cookpa:
How strange. Thanks for this, I'll try it out

cookpa:
Works for me

ntustison:
Thanks @cookpa .

PR #1574, opened 2023-07-25
Remove ANTSPATH

### LINKED ISSUE #1573: Do we still need ANTSPATH?
With ANTs now available for Conda, I was wondering whether there is any use for `ANTSPATH`?

Right now, some bash scripts cannot be run without explicitly setting `ANTSPATH` to the Conda env bin directory, despite the guarantee that all binaries will be present under `PATH`. I am actually wondering what use cases for `ANTSPATH` are not already covered by `PATH` overrides? 

I am tempted to patch all scripts to strip `ANTSPATH` from the Conda package, but it looks like other distributions of ANTs, like Homebrew, Debian or RPM packages, could benefit from this change too.

WDYT?

ghisvail:
Closes #1573

### DIFF
--- a/Scripts/ants.sh
+++ b/Scripts/ants.sh
@@ -3,12 +3,10 @@
 NUMPARAMS=$#
 
 MAXITERATIONS=30x90x20
-if [[ -z ${ANTSPATH} ]] ; then
-  echo "Environment variable ANTSPATH must be defined"
-  exit 1
-fi
-if [[ ! -f "${ANTSPATH}/ANTS" ]] ; then
-  echo "Cannot find the ANTS program. Please \(re\)define \$ANTSPATH in your environment."
+
+if ! command -v "ANTS" &> /dev/null
+then
+  echo "Cannot find the ANTS program. Please \(re\)define \$PATH in your environment."
   exit 1
 fi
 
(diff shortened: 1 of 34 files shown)


ghisvail:
Looks like I am missing a few files (32 vs 34 in the search).

cookpa:
Most of the scripts still need some work - removing ANTSPATH references is not sufficient because many of the checks for programs will fail. 

Eg

`-s "${ANTSPATH}ANTS"`

can't be replaced with

`-s "ANTS"`

but rather needs to be `-s $(which ANTS)`.

Is it OK if I push to your branch?

ghisvail:
Sure, go for it 👍

Le mar. 25 juil. 2023, 13:43, Philip Cook ***@***.***> a
écrit :

ghisvail:
> POSIX check for a command is `command -v` as which may not always be available

PR updated with `if ! command -v $CMD &> /dev/null` guards as per this SO [thread](https://stackoverflow.com/questions/592620/how-can-i-check-if-a-program-exists-from-a-bash-script).

Scripts/ants.sh
```diff
@@ -3,20 +3,17 @@
...
 
 MAXITERATIONS=30x90x20
-if [[ -z ${ANTSPATH} ]] ; then
-  echo "Environment variable ANTSPATH must be defined"
-  exit 1
-fi
-if [[ ! -f "${ANTSPATH}/ANTS" ]] ; then
-  echo "Cannot find the ANTS program. Please \(re\)define \$ANTSPATH in your environment."
+
+if ! command -v "ANTS" &> /dev/null
```
cookpa, line 7: I think you need the square brackets still

ghisvail: It works without brackets on my machine. I used the pattern in the accepted answer.

ghisvail:
```
# Make ANTS disappear from PATH.
$ mv $HOME/.local/ants/bin/ANTS{,.bak}

# Run wrapper script.
$ ./Scripts/ants.sh
Cannot find the ANTS program. Please \(re\)define $PATH in your environment.

# Put it back.
$ mv $HOME/.local/ants/bin/ANTS{.bak,}

# Run wrapper script again.
$ ./Scripts/ants.sh
 USAGE ::  
  sh   ants.sh  ImageDimension  fixed.ext  moving.ext  OPTIONAL-OUTPREFIX   OPTIONAL-max-iterations  OPTIONAL-Labels-In-Fixed-Image-Space-To-Deform-To-Moving-
```

cookpa:
Looking good now @ghisvail

cookpa:
Thanks for doing this @ghisvail

PR #1604, opened 2023-09-25
ENH: Allow priors to be specified with BIDS labels in ants*CorticalThickness.sh

cookpa:
Also allow templateflow labels (almost bids but use SCGM instead of SGM).

This lets users run ACT with a templateflow template, eg tpl-OASIS30ANTs, without first renaming the priors 01 through 06.

For backwards compatibility, and because Atropos requires it, output is still numbered 1-6.


ntustison:
Thanks for this @cookpa .

ntustison (approved):

PR #1636, opened 2023-12-14
ENH: Better to use actual index into imageset array

cookpa:
Name output files by position in input, rather than by tuple


cookpa:
The bug was fixed by #1635 because output is indexed by modality template as well as the file name, but this way the numbering always corresponds to the position of the image in the command line, including when number of modalities (-k) is > 1

PR #1647, opened 2024-01-05
ENH: Install cmake using apt

### LINKED ISSUE #1644: Minified, multi-architectural container images
**Is your feature request related to a problem?**

We are interested in using `N4BiasFieldCorrection` as part of our pipelines and run it in containers on PowerPC.

**Proposed new features**

A) In `Dockerfile`, by installing `cmake` using `apt` instead of directly from Kitware's website, we can build multi-architectural images using `docker buildx`. See here:

https://github.com/FNNDSC/ANTs/blob/04c2938830f6c1233bde9c4784ad2f58d6a8f058/Dockerfile

I intend to make a PR after figuring out a good solution to https://github.com/ANTsX/ANTs/issues/1069

B) Small (<60MB) images can be produced by creating individual images for each ANTs program. See

https://github.com/FNNDSC/microANTs

Using this script, I created a multi-architectural image for `N4BiasFieldCorrection` on DockerHub: https://hub.docker.com/layers/fnndsc/n4biasfieldcorrection/2.5.0/images/sha256-7e4da5a9f606aa170ae8d1187a6de915f5a1d03ca2dd7e773fc24c411c2227ab?context=explore (the ppc64le one doesn't work yet, as aforementioned)

If there is more interest in the community, I can use a GitHub Actions matrix to automatically build these minified images for all ANTs programs. I would also like to know whether the ANTsX team is interested in hosting these images and CI infrastructure in their own official namespace.

**Alternatives you've considered**

`docker buildx` is a very inefficient solution because it uses QEMU emulation. Cross-compilation would be a much more efficient solution, however it would require in-depth knowledge of the Cmake configurations of ANTs which I lack.

Alternatively, it would also be even better if the `ants` package in `conda-forge` was cross-compiled for linux-aarch64 and linux-ppc64le.

**Additional context**

jennydaman:
# Description

- In `Dockerfile`, the distro is switched to `debian:bookworm-slim` from `ubuntu:bionic`
- `cmake` is installed from Debian's package repositories, providing a solution which can be more easilly built on linux/arm64
- Use OCI spec annotations e.g. `org.opencontainers.image.title`, see https://github.com/opencontainers/image-spec/blob/main/annotations.md

Resolves #1644


cookpa:
Thanks for this @jennydaman

Can you add the `bc` tool to the final layer? Some of the ANTs scripts need that

cookpa:
I'm going to test this a bit before merging just to ensure there main scripts work with the new Linux distro

PR #1650, opened 2024-01-11
COMP: Fix some compile errors when legacy functionality is removed

dzenanz:
Legacy removal is forced to ON [when wrapping is enabled](https://github.com/InsightSoftwareConsortium/ITK/blob/c8ec91c826898838f5f1a1a2a53849dd3efc9c88/CMakeLists.txt#L305-L310). Prepare ANTs for building in such a scenario. We will need it soon in https://github.com/thewtex/ANTsWasm.


ntustison:
Thanks @dzenanz

PR #1651, opened 2024-01-11
A variety of compile fixes which appear when legacy is disabled

dzenanz:
Assorted compile fixes and update to the latest ITK. This is a follow-up to #1650.


cookpa:
Thanks for these updates @dzenanz

dzenanz:
This finally builds cleanly.

cookpa:
Awesome, thanks!

ntustison:
Yes, thank you very much @dzenanz .

Issue #1627, opened 2023-11-17
Atropos fails on macOS with ANTs 2.5.0


devDonnn:
Hi,

I've been running ANTs Atropos using the downloaded GitHub release binary ants-2.5.0-macos-12-X64-clang.zip for my Mac Studio, Apple M1 Max, macOS 14.1.1:
```
noglob Atropos -d 3 -x brainmask.nii.gz \
-c [ 200,0.0005 ] \
-a diff.nii.gz \
-i kmeans[ 2 ] \
-s 1x2 \
-k HistogramParzenWindows[ 1,32 ] \
-m [ 0.1,1x1x1 ] \
-o [ tmpSegmentation.nii.gz,posteriors%d.nii.gz ] \
-r 1 -p Socrates[ 0 ]
```

However, it crashes with the following terminal output:

> 
> Atropos(6030,0x30f7b5000) malloc: *** error for object 0x600002380f00: pointer being freed was not allocated
> Atropos(6030,0x30f7b5000) malloc: *** set a breakpoint in malloc_error_break to debug
> zsh: abort      noglob Atropos -d 3 -x VISIT01/T1_brainmask.nii.gz -c [ 200,0.0005 ] -a  -i

On the other hand, I had no problems with the ANTs 2.4.2 version (also downloaded GitHub release binary).

cookpa:
Hi @devDonnn , sorry for the delay in getting to this. I don't have an arm64 machine so I didn't investigate right away.

I can reproduce this though on Intel Mac and Linux. It goes away if I replace the likelihood model 'HistogramParzenWindows' with 'Gaussian'

Reproduced with

```
Atropos -d 3 -x tpl-MNI152NLin2009cAsym_res-02_desc-brain_mask.nii.gz \
-c [ 200,0.0005 ] \
-a tpl-MNI152NLin2009cAsym_res-02_T1w.nii.gz \
-i kmeans[ 2 ] \
-k HistogramParzenWindows[ 1, 32 ] \
-m [ 0.1,1x1x1 ] \
-o [ tmpSegmentation.nii.gz,posteriors%d.nii.gz ] \
-r 1 -p Socrates[ 0 ] \
--verbose
```

cookpa:
Many ANTs compilations later, I have isolated the commit:

```
0ac02f6f4475ce3d83f871d3437d512184b51cde is the first bad commit
commit 0ac02f6f4475ce3d83f871d3437d512184b51cde
[...]

PR #1663, opened 2024-01-22
BUG: Stop segmentation fault when Atropos is multi-threaded

cookpa:
Partial fix for #1627.

@ntustison, I tried replacing the local interpolator with a vector of interpolators for each bin at the class level. It improves the performance issue. I initialize the interpolators in the `SetInputListSample` function, which I think will cause them to stay up to date with respect to the m_HistogramImages, but I'm not sure.


ntustison:
Yeah, that should work.  Thanks @cookpa .

cookpa:
OK thanks @ntustison - I'm going to try to debug/fix the ManifoldParzenWindows model as well, then will merge

cookpa:
Tested some more, this actually fixes #1627 , the ManifoldParzenWindows error looks to be unrelated to the multi-threading changes

Issue #1587, opened 2023-08-15
Build/Install Defaults & Binaries -- Static vs Shared


gdevenyi:
The current default build process is `BUILD_SHARED_LIBS=OFF` which means ANTs binaries are statically built. Despite this, `make install` will still install all the libraries, which makes the install huge.

For docker containers, binaries, etc, we should either use the shared build (library paths should be handled by RPATH) or not install the libraries during static install.

cookpa:
The Dockerfile does use the shared builds, but I don't think it uses RPATH, I have to set `LD_LIBRARY_PATH`. This breaks shared Mac OS builds, because the OS does not respect `LD_LIBRARY_PATH` for security reasons.

I don't know if not installing the libraries could affect other ANTsX software, but would need to verify that before making any changes. Perhaps it could be an option, I believe we have a "libs only" option already, I could add "binaries only".

gdevenyi:
Related, https://github.com/ANTsX/ANTsPy/pull/346

PR #1665, opened 2024-01-23
ENH: Allow static libs, re-enable testing

cookpa:
Run the short tests during the Docker build, add an option to turn off shared libs

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -21,6 +21,3 @@ bin
 build
 *.orig
 *.bak
-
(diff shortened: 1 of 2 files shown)


gdevenyi:
Maybe address https://github.com/ANTsX/ANTs/issues/1587 ?

cookpa:
I would like to address #1587 at the CMake level so that binaries only can be installed, at which point it can become a docker option also

PR #1682, opened 2024-02-26
ENH: Support averaging affine transforms with float parameters

dzenanz:
So far, this was only possible with double parameters.

### DIFF
--- a/Utilities/itkAverageAffineTransformFunction.h
+++ b/Utilities/itkAverageAffineTransformFunction.h
@@ -55,7 +55,7 @@ struct HelperCommonType
                                   unsigned int                iEnd);
 };
 
-template <typename T>
+template <typename T, typename TParametersValueType>
 class HelperType;
 
 // {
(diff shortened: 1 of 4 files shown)


dzenanz:
Review/merge? This builds locally, and does not introduce any new test failures.

cookpa:
Looks good to me, have you confirmed that AverageAffineTransform / AverageAffineTransformNoRigid work as before? These are critical to the template building scripts, and I don't know if they are covered by the automated tests.

cookpa:
I'm building now and will test today

dzenanz:
The changes changes only add a new way to use the class, they do not change how class behaves.

Trying to use it with float affine matrices caused compile error. The new code path is untested for now. But it should be tested soon by unit tests in https://github.com/dzenanz/ANTsWasm. Of course, adding some tests for the float affine averaging within this repository would be good, but I have no plans to do that. Anyone is welcome to do so.

cookpa:
Thanks, merging

PR #1685, opened 2024-02-28
ENH: Retain docker image from PRs

cookpa:
This will make it easier to quickly download and test PRs, using temporary Github artifacts that expire after 7 days.


cookpa:
Note Mac Finder "decompress archive" tries to be extra helpful by unpacking the tar file as well, use the CLI unzip to avoid this.

```
unzip ants_image.tar.zip
docker load -i ants_image.tar
```

PR #1694, opened 2024-03-10
ENH: Check input for antsApplyTransforms matches dimensionality

cookpa:
This checks that scalar input to `antsApplyTransforms` is actually an N-dimensional scalar image, and if it's not, exit with error advising users to use the correct input flag with `-e`.

It's my opinion that it's wise to inform users of fatal errors even if they don't use `--verbose`, but if that's going to break anything I can hide the error message for non-verbose execution.

### DIFF
--- a/Examples/antsApplyTransforms.cxx
+++ b/Examples/antsApplyTransforms.cxx
@@ -266,9 +266,9 @@ antsApplyTransforms(itk::ants::CommandLineParser::Pointer & parser, unsigned int
       // can only check files, so skip checks if using a pointer in ANTsR / ANTsPy
       if (verbose)
       {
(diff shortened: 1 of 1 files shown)


cookpa:
Just noticed this elsewhere in the code. Might have to adapt the checks to only operate on actual files

```
  // BA - code below creates problems in ANTsR
  //  itk::ImageIOBase::Pointer imageIO = itk::ImageIOFactory::CreateImageIO(
  //                                                            filename.c_str(),
  //                                                            itk::ImageIOFactory::FileModeType::ReadMode );
  //  dimension = imageIO->GetNumberOfDimensions();
```

PR #1699, opened 2024-03-18
BUG:  Need to fill buffer for R and Python interfaces.

ntustison:
https://github.com/ANTsX/ANTsPy/issues/579

### DIFF
--- a/Utilities/itkDiReCTImageFilter.hxx
+++ b/Utilities/itkDiReCTImageFilter.hxx
@@ -259,19 +259,23 @@ DiReCTImageFilter<TInputImage, TOutputImage>::GenerateData()
   {
     CumulativeVelocityFieldPointer forwardCumulativeVelocityField = this->GetForwardCumulativeVelocityField();
     forwardCumulativeVelocityField->Allocate();
+    forwardCumulativeVelocityField->FillBuffer(zeroVector);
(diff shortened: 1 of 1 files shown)


stnava:
my understanding ( from itk discussions ) is the C++ standard does not define expectations for values within uninitialized memory; such values are indeterminate, and accessing them leads to undefined behavior.  ... not sure if itk Allocate is supposed to behave differently though

gdevenyi:
To fix this common issue itk has been transitioning to a single call for both

https://github.com/InsightSoftwareConsortium/ITK/pull/4494

cookpa:
Thanks @gdevenyi - so it looks like we could replace Allocate() with Allocate(true) or, as of that PR, AllocateInitialized() and solve this problem consistently.

stnava:
yes - worthwhile.  would also need to check all the R c++ and python c++ ( not much in the latter , I think ) .... another issue is I have not updated pybind11 in some time .... not sure if this would be related or not.

ntustison:
I'll go ahead and merge this and a future pull can address the more general presence of this issue.

ntustison:
Unless somebody else wants to do it, I can put it on my to-do list to go through and check that the image buffer is initialized.

cookpa:
Thanks Nick, much appreciated. I'll make some issues so we can keep track of this

ntustison:
Thanks @cookpa .  But just to be clear, it might take a couple days to a week.

Issue #1701, opened 2024-03-18
To do: Check all allocated memory is initialized


cookpa:
Relating to #1699, which addresses this issue for KellyKapowski

cookpa:
Fixed via #1702

PR #1702, opened 2024-03-18
ENH:  Check for initialization after image allocation.

ntustison:
https://github.com/ANTsX/ANTs/issues/1701

Hey @cookpa , it wasn't as much as I was expecting.


cookpa:
Thanks very much for this!

PR #1705, opened 2024-03-19
COMP: Cancel previous ci builds

cookpa:
If multiple pushes are made to the same branch or PR, cancel prior CI runs

Avoids wasting time and also prevents dockerhub getting out of sync


cookpa:
This will stop multiple builds sitting in the queue, but unfortunately there's no way to stop the cancellation emails

PR #1712, opened 2024-04-02
Refactor antsTransformInfo to add some error handling and compute determinant

gdevenyi:
I'd like to be able to easily extract the determinant of affine transforms and antsTransformInfo seems like the right place.

The function has been refactored to handle reading errors.

Based primarily off
https://github.com/InsightSoftwareConsortium/ITK/blob/master/Examples/IO/TransformReadWrite.cxx

Things that aren't done:
- I don't know how to automatically handle the 2D vs 3D transforms. The previous simple read/print code didn't seem to need it whereas this version does.

### DIFF
--- a/Examples/antsTransformInfo.cxx
+++ b/Examples/antsTransformInfo.cxx
@@ -19,6 +19,9 @@
 
 #include "antsCommandLineParser.h"
 #include "itkTransformFileReader.h"
+#include "itkMatrixOffsetTransformBase.h"
+#include "itkCompositeTransform.h"
+
 
 #include <sstream>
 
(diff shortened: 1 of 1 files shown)


cookpa:
Once the transform is read, I think you can call `GetInputSpaceDimension()` on it to determine the dimensionality

gdevenyi:
> The program as written is not restricted to affine transforms,

Can you give an example? I tried this on Warpfields and it falls over, I don't know of any other transform types which couldn't be cast to an affine...

cookpa:
If you run `antsRegistration` with `--output-composite-transform`, it will read the resulting transform file.

gdevenyi:
Got it.

Looks like I need to read all transforms as composite transforms and then figure out how to unpack it back into a list and loop over the unpacked components, checking the type.

cookpa:
From the discourse, seems like dynamic casting to MatrixOffsetTransformBase would be sufficient. If that's not null, you can call GetMatrix() on the result

gdevenyi:
This now will unpack transform lists as well as CompositeTransforms and print the determinant in addition to the PrintSelf for every transform in the whole stack.

cookpa:
Thanks - I'm going to be out of the office, so might be some time before I get to this

gdevenyi:
Context of this work is I use affine determinants in steps of
https://github.com/CoBrALab/optimized_antsMultivariateTemplateConstruction

And currently I have to do this by using antsApplyTransforms to produce a warp field from the affine and then a Jacobian file from that. This will allow me to just compute the the jacobian determinant with antsTransformInfo and extract it from the text output.

cookpa:
I didn't verify the determinant values but it appears to work with the supported transform types

gdevenyi:
I have tested the output values using MINC affine transforms, where the I can generate arbitrary transforms with `param2xfm` (giving a known determinant) and then compute the determinant using both MINC tools and antsTransformInfo :+1:

PR #1720, opened 2024-04-11
BUG: Should not assume DIM=3

### LINKED ISSUE #1718: ANTs apply transform for template construction has hardcoded dimension = 3
in the files antsMultivariateTemplateConstruction.sh and antsMultivariateTemplateConstruction2.sh, the calls to antsAI and antsApplyTransforms have hardcoded -d 3, when they should instead be -d ${DIM}. This was causing errors when trying to construct 2D templates.

cookpa:
Fixes #1718


cookpa:
Never mind

Issue #1733, opened 2024-04-30
Deprecation of LabelGeometryImageFilter (used in LabelGeometryMeasures)


cookpa:
https://github.com/InsightSoftwareConsortium/ITK/pull/4630

Thanks @gdevenyi for the heads up.

I don't use LabelGeometryMeasures much because it doesn't do what most users want, which is estimate cortical surface area.

I was actually looking at this issue recently. If I understand this paper correctly,

https://onlinelibrary.wiley.com/doi/abs/10.1002/1361-6374%28199303%291%3A1%3C6%3A%3AAID-BIO3%3E3.0.CO%3B2-3

it's possible to approximate surface area from a voxel image, if the local geometry is known. What we need though is a filter that computes this only for voxels connected to background.

stnava:
hmm - this would be a problem for me.   @cookpa  thickness = Volume / area  .... so we compute area on the fly when we need to ... A is the surface area of one side of the sheet.

cookpa:
Looks like this is a suggested replacement

https://itk.org/Doxygen/html/classitk_1_1LabelImageToShapeLabelMapFilter.html

cookpa:
I'll work on it, and do some comparisons on the results

gdevenyi:
I'm happy to test as I use LabelGeometryMeasures for some work.

cookpa:
As well as LabelGeometryMeasures, this filter is used in

```
ImageSegmentation/antsAtroposSegmentationImageFilter.hxx
Examples/ImageMath_Templates.hxx
Examples/GetConnectedComponentsFeatureImages.cxx
Examples/CreateDTICohort.cxx
```

cookpa:
LabelGeometryImageFilter is used in the undocumented "LabelThickness2" function in ImageMath. It appears to compute average thickness based on the ratio of the volume to perimeter, then replaces the input labels with the average thickness for each label.

@ntustison @stnava do we still need this?

https://github.com/ANTsX/ANTs/blob/4c68a48ea4f77f11197182eacabb74ce84749327/Examples/ImageMath_Templates.hxx#L10782-L10867

cookpa:
It seems that the pseudo-thickness image from ImageMath LabelThickness2 should be provided by `GetConnectedComponentsFeatureImages` - only there's a bug in the latter and it outputs area / volume, not volume / area.

ntustison:
[...]

PR #1737, opened 2024-05-02
ENH: Update LabelGeometryMeasures to not use deprecated filter

cookpa:
Not a fix but part of the way to fixing #1733 

I replaced the filter and added some new features including computing mean and median intensities of gray images.

Some attributes don't have an exact analog. I changed some headers to be general over 2D / 3D

New columns are

Label
NumberOfVoxels
PhysicalSize - area in 2D (mm^2) or volume in 3D (mm^3)
Perimeter - perimeter in 2D or surface area in 3D
Elongation - Note this differs from old implementation
Roundness - new
Flatness - new
Centroid_x - Note this differs from old implementation
Centroid_y
Centroid_z
BoundingBoxLower_x - reordered (open to changing this)
BoundingBoxLower_y
BoundingBoxLower_z
BoundingBoxUpper_x
BoundingBoxUpper_y
BoundingBoxUpper_z
MeanIntensity - Grayscale stats if intensity image provided
SigmaIntensity
MedianIntensity
MinIntensity
MaxIntensity

Eccentricity, Orientation have been removed as they don't exist in the new filter. Also, I didn't add the "WeightedCentroid" for gray images, I don't think that exists in `LabelStatisticsImageFilter`.

To check the centroid, I compared to SimpleITK:

```
#!/usr/bin/env python

import SimpleITK as sitk
import sys

def main():
    if len(sys.argv) < 2:
        print("Usage:", sys.argv[0], "<InputImage>")
        return

    # Read the image file
    image_path = sys.argv[1]
    image = sitk.ReadImage(image_path)

    # Convert to a binary image, assuming the image might not be perfectly binary
    binary_image = image > 0

    # Calculate the centroid using the LabelShapeStatisticsImageFilter
    label_shape_filter = sitk.LabelShapeStatisticsImageFilter()
    label_shape_filter.Execute(binary_image)

    # Assuming the label of interest is '1' (common for binary images)
    centroid = label_shape_filter.GetCentroid(1)
    print(f"Centroid of the binary object: {centroid}")

if __name__ == "__main__":
    main()
```

which agrees with my definition but note that it differs from the old filter and also what c3d reports (I am not sure what filter it uses)

### DIFF
--- a/Examples/LabelGeometryMeasures.cxx
+++ b/Examples/LabelGeometryMeasures.cxx
@@ -110,7 +110,7 @@ LabelGeometryMeasures(int argc, char * argv[])
 
   using FilterType = itk::LabelImageToShapeLabelMapFilter<LabelImageType>;
   typename FilterType::Pointer filter = FilterType::New();
-  filter->SetComputeOrientedBoundingBox(true);
+  filter->SetComputeOrientedBoundingBox(false);
   filter->SetComputePerimeter(true);
   filter->SetComputeFeretDiameter(false); // slow for large labels eg brain mask
   filter->SetInput(labelImage);
@@ -214,7 +214,7 @@ LabelGeometryMeasures(int argc, char * argv[])
     for (unsigned int idx = 0; idx < ImageDimension; ++idx)
     {
       if (principalMoments[idx] > 0)
-        axesLengths[idx] = 2.0 * std::sqrt(principalMoments[idx]);
+        axesLengths[idx] = 4.0 * std::sqrt(principalMoments[idx]);
       else
         axesLengths[idx] = 0.0;
     }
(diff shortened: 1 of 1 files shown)


ntustison:
Awesome.  Thanks @cookpa

cookpa:
Can't convert to draft from my phone but don't merge yet I will make a couple of changes to maintain backwards compatibility in column names as requested by @stnava

cookpa:
In the latest commit, I've added "Eccentricity" and "Axes Length" back, both derived from the label principal moments.

Axes length agrees with the old definition, which suggests that the moments are the same. But eccentricity differs.

cookpa:
More discussion of the numbers here

https://github.com/InsightSoftwareConsortium/ITK/pull/4624#issuecomment-2094818795

PR #1750, opened 2024-05-19
ENH: update and reformat README

ncullen93:
This proposed PR updates and reformats the README. It is best to view it visually on my fork at https://github.com/ncullen93/ANTs instead of the raw file. 

Improvements:
- You don't have to leave the readme to learn how to install ANTs in a basic scenario.
- All learning resources are grouped together so when I’m in “learning mode” I can quickly evaluate which one I need.
- Same with scripts / example code that are now grouped a bit more semantically. 
- For references, article titles are displayed so I can quickly evaluate which are relevant rather than having to follow a ton of links that take me away from the readme.
- In general, it's much clearer now when you're nevigating to a script, to a tutorial, or to a journal article. It was a complete lottery what I’m going to get from the links on the current Readme.
- In general, links that take users away from the repo should be minimized… they’ll get distracted and never come back.
- Links to outdated (e.g., source forge) or unfinished resources are all removed

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Advanced Normalization Tools (ANTs) is a C++ library available through the comma
 
 The ANTs library is considered a state-of-the-art medical image registration and segmentation toolkit which depends on the Insight ToolKit, a widely used medical image processing library to which ANTs developers contribute. ANTs-related tools have also won several international, unbiased competitions such as MICCAI, BRATS, and STACOM.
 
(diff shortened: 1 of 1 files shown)


ntustison:
Hey @ncullen93 , I think that looks very good.  I would obviously update the team list to include yourself and probably level the listing so that everybody's in the bulleted list or in a single sentence.  I'd probably also add a recent paper but I can do that after your merge.

stnava:
thanks for this 

perhaps link to the wiki some where [https://github.com/ANTsX/ANTs/wiki](https://github.com/ANTsX/ANTs/wiki)

ncullen93:
@ntustison I updated the team section and agree it would be nice for you to update or add references eventually.

@stnava Great idea - the ANTs Wiki is now the first link under the "Learning Resources" section.

Feel free to give more feedback or merge and then make any changes you want.

cookpa:
Merging in anticipation of next release (ITK 5.4.0 is about to be announced)

PR #1751, opened 2024-05-21
ENH: Add Ubuntu 24.04

gdevenyi:
Add 24.04 by copy-paste of the existing GH action.

### DIFF
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -16,6 +16,14 @@ jobs:
       matrix:
         config:
         - {
+            name: "Ubuntu-24.04-GCC",
+            os: ubuntu-24.04,
+            cc: "gcc",
+            cxx: "g++",
+            build_type: "Release",
+            generators: "Ninja"
+          }
+        - {
             name: "Ubuntu-22.04-GCC",
             os: ubuntu-22.04,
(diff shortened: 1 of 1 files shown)


cookpa:
Thanks, happy to add this, have you verified that it builds?

gdevenyi:
I was expecting it to build here in the PR, but I guess it won't since its a release-only trigger. I'm trying to figure out how to do it manually on my fork now...

cookpa:
It is annoying and I probably should make a workflow_dispatch option. The way I currently test is to make releases on my fork.

cookpa:
You know, I will just test it, I have it all set up

gdevenyi:
>  The way I currently test is to make releases on my fork.

I figured, I just tried that on my fork and it didn't work for me :-/

cookpa:
Yeah, you have to enable actions, and then disable the ones you don't want...I will fix this because things break on release and it would be good to be able to test

gdevenyi:
How fast are the builds? I was looking at adding this
https://github.com/marketplace/actions/ccache-for-gh-actions

But like this I can't really test it easily...

cookpa:
About an hour. I've used the cache before in python, it can help but caches only last a week so they are good for a nightly build but not really for releases

PR #1771, opened 2024-07-15
ENH: write template output such that things sort together

cookpa:
A few fixes

* Replaced all the "let" statements and other traps that prevent running with `bash -e`, and turned that on.

* Fixed error checking in pairwise registration stage

* Attempted to make output more readable by prefixing input indices, rather than adding them at the end. Trying to make things easier to list and review, especially with multiple modalities


gdevenyi:
If you're willing to go through some painful testing, this is the gold standard,
http://redsymbol.net/articles/unofficial-bash-strict-mode/

And even more painful, getting to pass, https://github.com/koalaman/shellcheck

cookpa:
Thanks @gdevenyi

Issue #1681, opened 2024-02-22
antsApplyTransforms not able to read transform txt


butellyn:
**Describe the problem**

I am trying to use `antsApplyTransforms` to move a functional image in fmriprep's T1w space into fsnative (freesurfer's T1w space), but the output is not being created. With the "verbose" option turned on, I got a message that the transform file cannot be read. The file was created using fMRIPrep v23.1.4. I am thinking maybe the version of ANTs on my cluster (which seems to be from 2020) may be too old to be compatible with the outputs from fMRIPrep. Does that seem likely to be the case? If not, what might the issue be?

**To Reproduce**

I ran this command (which took ~30 seconds):
```
antsApplyTransforms -d 4 -v 1 \
    -i /projects/b1108/studies/mwmh/data/processed/neuroimaging/fmriprep_23.1.4/sub-MWMH212/ses-2/func/sub-MWMH212_ses-2_task-rest_space-T1w_desc-preproc_bold.nii.gz \
    -r /projects/b1108/studies/mwmh/data/processed/neuroimaging/surf/sub-MWMH212/ses-2/anat/fs_T1w.nii.gz \
    -t /projects/b1108/studies/mwmh/data/processed/neuroimaging/fmriprep_23.1.4/sub-MWMH212/ses-2/anat/sub-MWMH212_ses-2_from-T1w_to-fsnative_mode-image_xfm.txt \
    -o /projects/b1108/studies/mwmh/data/processed/neuroimaging/surf/sub-MWMH212/ses-2/func/sub-MWMH212_ses-2_task-rest_space-fsnative_desc-preproc_bold.nii.gz
```

It returned no output:
<img width="1228" alt="Screenshot 2024-02-22 at 3 49 19 PM" src="https://github.com/ANTsX/ANTs/assets/41019524/eda70af9-7aa7-4a54-abfc-ad8d3d9423bf">

And said that it could not read the initial transform:
<img width="1227" alt="Screenshot 2024-02-22 at 3 59 20 PM" src="https://github.com/ANTsX/ANTs/assets/41019524/eac30e90-a193-4c50-a152-81fe076162f2">

But the file exists, and is output from fMRIPrep v23.1.4: 
<img width="1234" alt="Screenshot 2024-02-22 at 4 00 11 PM" src="https://github.com/ANTsX/ANTs/assets/41019524/22fbd1a4-9ac3-4975-80da-51729780021d">

[...]

PR #1800, opened 2024-10-02
ENH: apply transform to single slice of time series

### LINKED ISSUE #1555: Enable antsApplyTransforms to select one volume in a 4D file to resample
**Is your feature request related to a problem?**

To apply a concatenated sequence of transforms to each volume in a 4D series (for example, head motion + susceptibility warp + coregistration), either the transforms must be the same for each volume or you must split the series and apply the desired transforms.

**Proposed new features**
I suggest adding either a flag or syntax for specifying specific volumes of 4D input. For example using array notation or an `--index` flag:

```
antsApplyTransforms -i raw_bold.nii[0] -r mni.nii -o mni_vol0.nii \
    --transform vol0_motion.txt \
    --transform susceptibility_warp.h5 \
    --transform coreg.txt
```

```
antsApplyTransforms -i raw_bold.nii --index 0 -r mni.nii -o mni_vol0.nii \
    --transform vol0_motion.txt \
    --transform susceptibility_warp.h5 \
    --transform coreg.txt
```

**Alternatives you've considered**
The current approach is to split the series into individual volumes, apply the transformations, and then concatenate the resulting series. It works, but generates 2N volumes for an N time point series.

An ideal approach would be to be able to specify a per-volume transform as one in a sequence of transforms such as

```
antsApplyTransforms -i raw_bold.nii -r mni.nii -o mni_bold.nii \
    --transform [vol0_motion.txt, vol1_motion.txt, ...] \
    --transform susceptibility_warp.h5 \
    --transform coreg.txt
```

However, that seems like a more significant lift. Reading from one volume in a series will at least cut the number of files generated in half, which would be helpful in itself.

**Additional context**

As mentioned above, splitting files before applying transforms creates 2N output files. In the context of a nipype pipeline, add in an additional factor of 5-10 inodes for working directories and their contents. This can become problematic when running a pipeline on many subjects on shared network filesystems.

cookpa:
A common use case is to register a time series to a scalar reference, then apply the warp to each volume. Sometimes, users will have different transforms for each volume (say, motion correction in fMRI). Other times, the time series might be very large after resampling into the fixed space.

This commit adds an option --time-index, which allows the user to select a particular volume. This is then extracted and transformed as a scalar image.

Fixes #1555

### DIFF
--- a/Examples/antsApplyTransforms.cxx
+++ b/Examples/antsApplyTransforms.cxx
@@ -210,6 +210,18 @@ antsApplyTransforms(itk::ants::CommandLineParser::Pointer & parser, unsigned int
   typename itk::ants::CommandLineParser::OptionType::Pointer inputOption = parser->GetOption("input");
   typename itk::ants::CommandLineParser::OptionType::Pointer outputOption = parser->GetOption("output");
 
+  // Time-index to extract from time-series image
+  unsigned long extractTimeIndex = 0;
+
+  bool extractTimeIndexSet = false;
+
+  typename itk::ants::CommandLineParser::OptionType::Pointer timeIndexOption = parser->GetOption("time-index");
+  if (timeIndexOption && timeIndexOption->GetNumberOfFunctions())
+  {
+    extractTimeIndexSet = true;
+    extractTimeIndex = parser->Convert<unsigned int>(timeIndexOption->GetFunction(0)->GetName());
+  }
+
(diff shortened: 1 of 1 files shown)


cookpa:
Relevant also to dealing with very large images like in #1681

ntustison (approved):

ntustison:
Thanks @cookpa

PR #1817, opened 2024-12-11
ENH: identity init transform for (bspline)syn-only

cookpa:
If using -t so or -t bo, the COM initialization is usually not wanted


ntustison:
THanks @cookpa

PR #1820, opened 2024-12-13
BUG: Fix segmentation fault for TranslationTransform in antsTransformInfo

### LINKED ISSUE #1819: antsTransformInfo fails to read COM alignment from antsAI
### Operating system and version

Mac OS 14

### CPU architecture

x86_64 (PC, Intel Mac, other Intel/AMD)

### ANTs code version

2.5.3.post16-g1892fef

### ANTs installation type

Compiled from source

### Summary of the problem

 Seg fault after using antsAI "AlignCentersOfMass"
 
 The transform is good, readable into antspy
 
```
 >>> import ants
tr = ants.read>>> tr = ants.read_transform('com.mat')
>>> tr
ANTsTransform
	 Type       : TranslationTransform
	 Dimension  : 3
	 Precision  : float
```	 

### Commands to reproduce the problem.

```
antsAI -d 3 \
                 -m Mattes[tpl-MNI152NLin2009cAsym_res-01_T1w.nii.gz, moving.nii.gz] \
                  -t AlignCentersOfMass \
                  -o com.mat \
                  --verbose 
 antsTransformInfo com.mat 
 ```
 

### Output of the command with verbose output.

Transform file: com.mat
Number of transforms = 1
zsh: segmentation fault  antsTransformInfo com.mat

### Data to reproduce the problem

Works on any data

gdevenyi:
Fixes #1819

### DIFF
--- a/Examples/antsTransformInfo.cxx
+++ b/Examples/antsTransformInfo.cxx
@@ -135,6 +135,9 @@ antsTransformInfo(std::vector<std::string> args, std::ostream * /*out_stream = n
             }
           }
         }
+        else if (!strcmp((*it)->GetNameOfClass(), "TranslationTransform")) {
+          it->Print(std::cout);
+        }
         else
         {
           using TransformType3D = itk::MatrixOffsetTransformBase<double, 3, 3>;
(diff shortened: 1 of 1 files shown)


cookpa:
I think this fix is good, but maybe it would be better to have antsAI write the translation as a `Euler3DTransform`, which is what `antsRegistration` does.

cookpa:
I'll see if I can change antsAI to bypass this - might solve the composition problem

cookpa:
Thanks for looking into this @gdevenyi, I will add a generic cast check into the else statement, but I don't think we need to handle Translation explicitly now

Issue #770, opened 2019-05-30
Atlas of the ANTS output, how to get its corresponding segmentation map.


taipingqu:
Dear ANTS expert, I have created a few brain atlas, I have tried the following command: "antsMultivariateTemplateConstruction2.sh -d 3 -i 2 -k 1 -f 6x4x2x1 -s 3x2x1x0 -q 100x100x70x20 -t SyN -m CC -c 0 -o 27_ ./*.nii", after outputting the template, but there is no segmentation data for the atlas. If I would like to get the segmentation data of the atlas, what command could I execute?

ntustison:
Segmentation is performed using the ``Atropos`` tool.  Here's a 2-D (example)[https://github.com/ntustison/antsAtroposN4Example] of how it's used in conjunction with bias correction.  For your 3-D data, you'll have to download one of the templates we've made available [here](https://figshare.com/articles/ANTs_ANTsR_Brain_Templates/915436) or [here](https://figshare.com/articles/ANTs_files_for_mni_icbm152_nlin_sym_09a/8061914).

PR #1831, opened 2025-01-24
ENH: Update default gradient step, turn off histogram matching

cookpa:
Have left antsBrainExtraction.sh alone because I'm not sure if histogram matching is more important where Laplacians are involved

Trying to harmonize with #768 and #770 in ANTsPy.

There are a lot of specialized / old scripts that I didn't change, we should consider moving to a permanent archive.

Histogram matching also defaults to on in antsMotionCorr, but again that is kind of a specialized case


cookpa:
Converting to draft because I need to test the scripts

PR #1832, opened 2025-01-26
STYLE: Add itkVirtualGetNameOfClassMacro + itkOverrideGetNameOfClassMacro

hjmjohnson:
Added two new macro's, intended to replace the old 'itkTypeMacro' and
'itkTypeMacroNoParent'.

The main aim is to be clearer about what those macro's do: add a virtual
'GetNameOfClass()' member function and override it. Unlike 'itkTypeMacro',
'itkOverrideGetNameOfClassMacro' does not have a 'superclass' parameter, as it
was not used anyway.

Note that originally 'itkTypeMacro' did not use its 'superclass' parameter
either, looking at commit 699b66cb04d410e555656828e8892107add38ccb, Will
Schroeder, June 27, 2001:
https://github.com/InsightSoftwareConsortium/ITK/blob/699b66cb04d410e555656828e8892107add38ccb/Code/Common/itkMacro.h#L331-L337

# Description

### DIFF
--- a/Examples/antsCommandIterationUpdate.h
+++ b/Examples/antsCommandIterationUpdate.h
@@ -125,7 +125,7 @@ public:
   /**
    * Run-time type information (and related methods).
    */
-  itkTypeMacro(antsCommandIterationUpdate, itk::Command);
+  itkOverrideGetNameOfClassMacro(antsCommandIterationUpdate);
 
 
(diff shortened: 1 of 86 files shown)


ntustison:
Thanks @hjmjohnson

PR #1839, opened 2025-02-03
Write composite transforms from antsApplyTransforms

cookpa:
antsApplyTransforms only allowed users to collapse linear transforms into a single .mat file, or composite transforms into a single displacement field.

Now allow users to write the full composite transform. This enables users familiar with `antsApplyTransforms` to write out composite transforms by specifying the components in the order they are applied when warping images. That is, one can call `antsApplyTransforms -o warpedImage` and then `antsApplyTransforms -o CompositeTransform[composite.h5]` without changing the `-t` options.

The motivation is to make it easier to write composite transforms, since CompositeTransformUtil requires users to use the internal "reverse queue order" of itkCompositeTransform.

The changes are backwards compatible, but I have noted in the usage that we prefer

`-o DisplacementField[ field.ext ]`

to

`-o [ field.ext, 1 ]`

Thus the output options are

```
-o warped.ext # apply transforms to input image
-o Linear[ collapsed.mat invert=0] # Write collapsed linear transform
-o DisplacementField[ collapsed.ext ] # Write collapsed displacement field
-o [ collapsed.ext, 1 ] # same as above, for backwards compatiblity
-o CompositeTransform[ composite.h5 ] # write composite transform
```

### DIFF
--- a/Examples/CompositeTransformUtil.cxx
+++ b/Examples/CompositeTransformUtil.cxx
@@ -40,7 +40,12 @@ PrintGenericUsageStatement()
             << " <transform name prefix>" << std::endl
             << "or" << std::endl
             << "CompositeTransformUtil  --assemble <CompositeTransform> "
-            << "<transform 1> <transform 2 > ... <transform N>" << std::endl;
+            << "<transform 1> <transform 2 > ... <transform N>" << std::endl
+            << std::endl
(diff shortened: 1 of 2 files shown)


cookpa:
Some tests 

https://github.com/cookpa/antsTransformTests

PR #1838, opened 2025-01-31
ENH: use prefix of 1st param from --output for --write-interval-volumes

mathdugre:
# Description
This PR changes the behavior of `--write-interval-volume` to use the prefix of the first parameter from `--output` as the write location, instead of `$PWD`.

Fixes #1836

PR #1840, opened 2025-02-04
ENH: use prefix from --output for --write-interval-volumes

mathdugre:
# Description
While doing more testing with the changes from #1838, I realized that `Examples/antsDisplacementAndVelocityFieldRegistrationCommandIterationUpdate.h` also needed to be changed to write interval volumes using the first prefix of `--output` when using SyN transform.

Related to #1838


cookpa:
Good catch, thanks

PR #1847, opened 2025-02-14
ENH:  Update ITK tag.

ntustison:
Hey @cookpa , it's been about a year since we've updated ITK.  I recently fixed an ITK bug that affects some tools in the ANTsX ecosystem and would like to reflect the ITK changes here.  Any potential issues that you're aware of?

### DIFF
--- a/SuperBuild/External_ITKv5.cmake
+++ b/SuperBuild/External_ITKv5.cmake
@@ -152,7 +152,7 @@ if(NOT DEFINED ${extProjName}_DIR AND NOT ${USE_SYSTEM_${extProjName}})
 
   ### --- End Project specific additions
   set(${proj}_REPOSITORY ${git_protocol}://github.com/InsightSoftwareConsortium/ITK.git)
-  set(${proj}_GIT_TAG 4535548a8539757c5fe9d81f8de5d804cd0a384f) # 2024-03-12
+  set(${proj}_GIT_TAG 018e0eab2bae8b377fd0ee3af83fdcbf22a36f93) # 2025-02-14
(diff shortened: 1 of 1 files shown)


cookpa:
I had hoped to build against the 5.4.2 release but it seems like that doesn't include your fix.

What tools are affected by this patch?

ntustison:
@thewtex [mentions](https://github.com/InsightSoftwareConsortium/ITK/pull/5242) backporting to ITK 5.4.3.  

The bug affects b-spline fitting but only for closed parametric dimensions.  So almost none of the tools in ANTsX would be effected except for FitBSplineToPoints.cxx and fitBsplineObjectToScatteredData in ANTsR (and its analog in ANTsPy).  If you want to hold off, I'm fine with that.

cookpa:
Thanks for the info. I think it's fine to update, we can update again when 5.4.3 comes out. It mostly helps the conda folks to have tagged versions of ants that rely on tagged versions of ITK.

ntustison:
Okay, thanks.  Let me finish compiling and testing locally and then I'll merge.

ntustison:
So follow-up question @cookpa---can we update the ANTs git tag in ANTsPy and ANTsR or should we hold off?

cookpa:
In ANTsPy we will need to update and verify the label stats code

https://github.com/ANTsX/ANTsPy/issues/612

I did what I could in #1737 to make this change less painful. The one thing I couldn't get from the new filter was the weighted centroid

cookpa:
ANTsR will need to update [labelGeometryMeasures.R](https://github.com/ANTsX/ANTsR/blob/5fb52b1616eb43866cc5bc53982006f71845ae08/R/labelGeometryMeasures.R)

ntustison:
Okay, I'll take a look.  Thanks.

PR #1849, opened 2025-02-17
BUG: CreateTiledMosaic would not write images without an RGB overlay

cookpa:
This was because it tried to write the mosaic with RealType pixels, which many formats do not support.

COMP: Bump ITK version string in cmake

### DIFF
--- a/Examples/CreateTiledMosaic.cxx
+++ b/Examples/CreateTiledMosaic.cxx
@@ -13,6 +13,7 @@
 #include "itkFlipImageFilter.h"
 #include "itkImageDuplicator.h"
 #include "itkImageRegionIterator.h"
+#include "itkIntensityWindowingImageFilter.h"
 #include "itkImageRegionConstIteratorWithIndex.h"
 #include "itkLabelStatisticsImageFilter.h"
 #include "itkPermuteAxesImageFilter.h"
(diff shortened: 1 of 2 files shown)


gdevenyi:
Can you give an example of a command this fixes? I've been fighting with trying to create mosiac images for a long time and I'm wondering if I had issues with this. Thanks.

cookpa:
```
CreateTiledMosaic -i t1w.nii.gz -o mosaic.png -t -1x-1 -p 5 -s 3 -d 2 -f 0x1
````

used to fail with something like

```
terminate called after throwing an instance of 'itk::ExceptionObject'
  what():  /build/ITKv5/Modules/IO/PNG/src/itkPNGImageIO.cxx:544:
PNG supports unsigned char and unsigned short
```

The way it's used in ANTs scripts always worked

```
CreateTiledMosaic -i t1w.nii.gz -o mosaic.png -r rgb.nii.gz -t -1x-1 -p 5 -a 0.5 -s 3 -d 2 -f 0x1
```

because the RGB overlay was used to compute an RGB pixel in the output, so the type conflict didn't happen.

gdevenyi:
Cheers, I already had sorted out a way to avoid this one when I don't have any overlay. This is nice to have solved.

PR #1852, opened 2025-02-24
ENH: more LabelGeometryMeasures fixes

cookpa:
# Description


cookpa:
NVM - I just didn't remove cached ants source in my antspy install

Issue #1860, opened 2025-03-21
antsAI local optimizations are not consistent with different search parameters


cookpa:
This is because the ITK optimizer does not reset its learning rate between start points. Discovered by @spinicist :

https://discourse.itk.org/t/unexpected-behaviour-in-multistartoptimizerv4/7466

ITK fix submitted:

https://github.com/InsightSoftwareConsortium/ITK/pull/5284

cookpa:
@ntustison I actually think this won't affect registration, because a scales estimator is already used in the linear stages where a conjugate gradient optimizer is used.

cookpa:
antsMotionCorr might be affected, but it has an option `-e 1` that will turn on scales estimation and avoid the bug.

cookpa:
Fixed by [#1861](https://github.com/ANTsX/ANTs/pull/1861)

PR #1861, opened 2025-03-21
BUG: Set scales estimator for antsAI

cookpa:
Prevents the learning rate being carried over to new start points

Should fix (#1860) without having to bump ITK

Somewhat inefficient but doesn't appear to have a large impact on run time (about 10% longer)


cookpa:
I did some more tests and it's not always slower, perhaps the benefit of fixing the step size leads to fewer line search iterations in some cases.

Issue #1137, opened 2021-01-25
Warp and reorient a diffusion tensor image issue with "large" affine transformation, looking for help/confirmation


valeryozenne:
Hi ANTs experts,

##

All data and scripts to reproduce the results is available at this link:
https://filesender.renater.fr/?s=download&token=491f1769-c437-426b-b210-51f2d65a0bf2
Data were down sampled to lower resolution to reduce the size of the files.

## Methods

I have been using ANTs for some months for reorienting tensors. I'm really happy to use it. The wikipage was really useful.
https://github.com/ANTsX/ANTs/wiki/Warp-and-reorient-a-diffusion-tensor-image
 
I still have one surprising result when combining transformations (affine + warp) in some cases (when the affine transformation is significant). I wonder if you could have a look.
I try to simplify as much the script, and to follow the convention and naming of the documentation : the script either apply on tensor images

* [COMPOSE MODE] : affine + warp transformations using the instruction for creating a compose transform in the wiki page 
* [STEP BY STEP MODE] : affine transformation, then reorient then apply the warp transformation and then reorient, here interpolation are done twice.

I have N datasets, here I shared only 2,  (see the results below)
* for dataset 1, [COMPOSE MODE] give wrongs results while [STEP BY STEP MODE] looks accurate, 
* for dataset 2  results are right for both mode. 
 
I'm suspecting either a mistake from my side or something related to the transformation itself, the affine transform make a significant shift. At the beginning, I also suspect the reference (fixed.nii.gz), or the physical space  but the result is accurate if I apply transformations step by step.
 
Would it be possible the process, applyTransform + ReorientTensorImage uisng compose transform doesn't work for some specific tranformations ?

My full script (not include here) include Rigid + Affine + Warp , leading to 3 interpolation in step by step mode to get the correct orientation, I wish to do it in one step using compose transform . 

Thanks in advance for your help. 
 
Valéry

## Results

I'm looking at the fiber orientation in the intraventricular septum, anatomy is simple, there are two distincts layers (in blue) fibers going up/down and (in green) fibers that are parallel to the tangent of the surface in the short axis view (transverse view). 
![Description](https://user-images.githubusercontent.com/7050315/105778409-aab49480-5f6c-11eb-8bd9-9937bb127061.png)

For dataset 2 , comparison of COMPOSE MODE is similar to STEP BY STEP MODE, the difference is due to the fact that interpolation is done twice.
[...]

PR #1871, opened 2025-04-30
BUG: Fix tensor reorientation with displacement fields

cookpa:
There was a lot of unused code in the PPD filter, and it was not consistent about how it expected the input tensors to be based - it seems it corrected for the header orientation with affine transforms, but not with displacement fields. A comment suggested this was intentional, but it produces wrong results if the resampled tensor image does not have an identity header transform. 

I've simplified the code in this PR.

Tensors should be stored in voxel orientation. In antsApplyTransforms, they are reoriented correctly into the voxel orientation of the reference space (#678).  In the new PPD filter, the tensors are rebased into physical orientation, the rotation from the transform is applied, then they are rebased onto the voxel orientation.

My motivation for getting back into this issue is the forthcoming BIDS extension proposal for dMRI derivatives. I plan to eventually do all the reorientation in `antsApplyTransforms`, but first I wanted to get the existing workflow correct. 

There remains the issue of how to import / export FSL tensors, these will require special handling, partly because of different NIFTI storage conventions, but also the tensor orientation in FSL has an x-axis flip compared to the ITK direction matrix, if the header transform has a positive determinant. This will require some special handling on input.

### DIFF
--- a/Examples/ImageMath.cxx
+++ b/Examples/ImageMath.cxx
@@ -328,6 +328,14 @@ ImageMath(std::vector<std::string> args, std::ostream * itkNotUsed(out_stream))
               << std::endl;
     std::cout << "    Usage        : ComponentTo3DTensor component_image_prefix[xx,xy,xz,yy,yz,zz] extension"
               << std::endl;
+    std::cout << "  FSLTensorToITK    : Converts a tensor image from FSL 4D format. Will flip the x-axis if the image "
+                 "is in neurological orientation."
+              << std::endl;
+    std::cout << "    Usage        : FSLTensorToITK 4D_DTImage.ext" << std::endl;
+    std::cout << "  ITKTensorToFSL    : Converts a tensor image to FSL 4D format. Will flip the x-axis if the "
+                 "image is in neurological orientation."
+              << std::endl;
+    std::cout << "    Usage        : ITKTensorToFSL 3D_DTImage.ext" << std::endl;
     std::cout << "  ExtractComponentFrom3DTensor    : Outputs a component images. " << std::endl;
     std::cout << "    Usage        : ExtractComponentFrom3DTensor dtImage.ext which={xx,xy,xz,yy,yz,zz}" << std::endl;
     std::cout << "  ExtractVectorComponent: Produces the WhichVec component of the vector " << std::endl;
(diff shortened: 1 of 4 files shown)


cookpa:
Added two commands to ImageMath:

FSLTensorToITK <input 4D tensor>
ITKTensorToFSL <input 3D tensor>

FSL requires radiological orientation of bvecs / tensors, regardless
of image orientation. This causes a problem for "NEUROLOGICAL" images
(positive determinant of direction matrix), because the tensors are then
not transformed to physical space by the direction matrix.

On import, we fix this by applying a reflection in x, and we do the same
on export back to FSL tensor format.

cookpa:
I think this PR may fix a longstanding issue #1137 - different behavior of ReorientTensorImage when given affine vs displacement fields. A few others have brought up similar issues, eg #1461.

I think the divergence between affine / displacement fields were that the affine transform was being corrected for the direction matrix, but the displacement field transform was not. I was confused by the `GetLocalDeformation` function, which does apply the image direction - this function was called in the filter, but then the results were not used. So in the end, displacement field reorientation was being applied to tensors still in the voxel orientation.

In this PR, I've tried to simplify things by rebasing the tensors to physical space (rather than modifying the transform), then applying the reorientation as before, then rebasing back to voxel space for output.

In short: reorientation with a displacement field was incorrect if the fixed image (into which the tensors are resampled by antsApplyTransforms) has a non-identity transform. The resampling itself is not affected; so tensor FA, MD etc are correct.

cookpa:
I think this is close to ready. There's still a systematic difference between mat and warp that is bigger than I'd like

image

this is "yawToYawRotated" in https://github.com/cookpa/antsDTOrientationTests.

The offset might be because the warp field is sampled coarsely (3mm). I'll investigate further.

Despite this, the major errors look to be resolved. I don't know how I didn't see these before. Lots of layers of confusion including the radiological / neurological display thing, and changing basis of the tensors (before #678), and limited scope of the bug. The three conditions for wrong results are:

1. There's a physical space rotation
2. The transform is a displacement field
3. The fixed image has a non-identity direction matrix. 

Merely rebasing the tensors into another image space with antsApplyTransforms was fine after #678.

cookpa:
I did some more digging, example scripts here

https://github.com/cookpa/itkTensorReorientation

The Jacobian estimation in the displacement field transform systematically underestimates rotation if the angle is large. When tested using rotations converted with TransformToDisplacementField, the estimated rotation is pretty good in the range 0-15 degrees, but by 30 degrees, it is consistently 5-6 degrees off.

I'm going to update the code to use a composite transform - this way large rigid rotations can be more accurately preserved.

PR #1879, opened 2025-05-07
Apply PPD reorientation for tensor data in antsApplyTransforms

cookpa:
Now that the PPD filter supports composite transforms, it is feasible to add it directly to antsApplyTransforms. This avoids the need to use a separate program and keep track of the transforms to use.

This simplifies things for users, and reduces potential for mistakes in transform composition.

To make this happen, I had to change the tensor type in AAT from a general itk::SymmetricSecondRankTensor to a DiffusionTensor3D. Typing this, I realize I should document this in the usage.

I think DiffusionTensor3D is the vast majority of use cases, so I'm thinking it would not be worth complicating usage by separately supporting generic tensors. Non-diffusion tensors might not also work with the log transformation that we apply automatically.

### DIFF
--- a/Examples/ImageMath.cxx
+++ b/Examples/ImageMath.cxx
@@ -322,7 +322,8 @@ ImageMath(std::vector<std::string> args, std::ostream * itkNotUsed(out_stream))
               << std::endl;
 
     std::cout << "\nTensor Operations:" << std::endl;
-    std::cout << "  4DTensorTo3DTensor    : Outputs a 3D_DT_Image with the same information. " << std::endl;
+    std::cout << "  4DTensorTo3DTensor    : Outputs a 3D tensor image from a 4D image containing upper-triangular components "
+              << "(dxx, dxy, dxz, dyy, dyz, dzz)" << std::endl;
     std::cout << "    Usage        : 4DTensorTo3DTensor 4D_DTImage.ext" << std::endl;
     std::cout << "  ComponentTo3DTensor    : Outputs a 3D_DT_Image with the same information as component images. "
               << std::endl;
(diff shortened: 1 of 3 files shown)


cookpa:
@ntustison @stnava will this interfere with any of your workflows? Should I make this not the default behavior? I could add a `--reorient` flag

ntustison:
I'm pretty sure this will be fine with anything I'm doing.  Thanks @cookpa.

cookpa:
Thanks Nick!

I remembered Brian had posted some code from antsPyMM over at

https://github.com/ANTsX/ANTsPy/issues/753

Looking at that, I see he is splitting the channels, warping, then reorienting. This would continue to work and would be the alternative I would recommend to use `antsApplyTransforms` without `-e 2`, if for some reason you didn't want to reorient.

stnava:
yeah - I think this change that you are making will not impact antspymm.   although I tested antspymm reorientation with a large variety of cases, am still not sure it's correct.   I'd be happy to switch to something that is more well validated if possible.

cookpa:
Thanks @stnava I think the antspymm method does have some of the same problems I have been working on. It is a really tricky thing to debug because the image headers, the code, and the visualization software can all change things. Will open an issue on antspymm.

I need to evaluate more, and am getting some help from a collaborator soonish.

cookpa:
I'm going to merge later unless I think of any other issues. I think this + other recent fixes solve the coordinate basis problem, and the new ImageMath edits allow for import of tensors from FSL, taking into account the change in sign of the x-axis for direction matrices with positive determinants.

There remains the question of how best to estimate the local rotation from a displacement field. I have a couple of prototype scripts that I will try to develop further, and test on real data somehow. I think the real test is doing whole-brain tractography

Issue #1881, opened 2025-05-09
Vector reorientation during transformation: Preservation of Only Direction (POD)?


cookpa:
I had a moment of panic when I thought about this, because the template script does

 ```
${WARP} -d ${dim} -e vector -i ${templatename}0warp.nii.gz -o ${templatename}0warp.nii.gz -t [ ${templatename}0GenericAffine.mat,1 ] -r ${template} --verbose 1
```

and I know warp field vectors are in physical space coordinates, while `antsApplyTransforms` assumes they are in moving voxel coordinates, and rebases them to the fixed image voxel coordinates.

However, in the shapeupdatetotemplate function, the moving and fixed direction matrices are always the same, so the rebasing cancels to identity.

Before the call above, the average affine transform is applied to the average warp. No reorientation step is done, so if there are rotations in the affine, the vector update will be off as a result.

This problem should converge to near zero after an iteration or two, especially if an affine-only template is built first.

But I wonder if it could contribute to suboptimal convergence if the first round is also deformable.

More generally, it would be nice to have vector reorientation so that we could reorient vectors as well as tensors. It could be like PPD except that it would renormalize to the original vector magnitude. I mostly foresee this being useful for directional data, like fiber orientations, though maybe it would help for templates too.

This would require adding yet another input type to antsApplyTransforms, because one would have to rebase to physical space to compute the orientation from the matrix.

PR #1895, opened 2025-07-01
ENH: Vector reorientation

cookpa:
Reorient vectors like we do for tensors. 

This composes the local jacobians from the composite transform into one, then does a polar decomposition to get a rotation. Thus it should preserve vector magnitude.

It's up to the user to ensure that the vectors are sensibly interpolated. For axial data, it probably makes sense to use nearest neighbor interpolation. At the least, data should be projected into one hemisphere so that we don't interpolate **x** and -**x**.

I've also added an extra option for displacement vector data in `antsApplyTransforms`. Displacement fields from ITK / ANTs registration are based in physical space. Generic vectors (-e 1) are assumed to be in index space, and are rebased accordingly.

For displacement fields, the vectors are signed and should have some spatial regularity, so linear interpolation might work better.

### DIFF
--- a/Examples/antsApplyTransforms.cxx
+++ b/Examples/antsApplyTransforms.cxx
@@ -1043,19 +1043,19 @@ antsApplyTransformsInitializeCommandLineOptions(itk::ants::CommandLineParser * p
 
   {
     std::string description = std::string("Option specifying the input image type of scalar (default), ") +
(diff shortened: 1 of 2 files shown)


cookpa:
Related to #1881

I've tested this vs DT orientation - if I pull V1 before applying the warp, I can get similar results to the tensor after PPD. 

Need to test on template building. The idea is if the mean affine is not close to zero, reorienting the warps will produce better shape updates.

Issue #397, opened 2017-01-05
bash scripts don't catch errors


gdevenyi:
If any of the commands in the bash scripts fail, currently the scripts continue on as if things were successful.

It would be beneificial to enable "unofficial bash strict mode" http://redsymbol.net/articles/unofficial-bash-strict-mode/ on the scripts to catch/error out on problems.

This *may* cause scripts to stop due to certain bash programming anti-patterns (which would need to be adjust) so it needs to be tested before being committed.

ntustison:
If you're interested in this, you're certainly welcome to test it and propose a relevant commit.

gdevenyi:
Sure. Can you point out which scripts you consider "maintained"?

Does there exist any small test data for these scripts so I can confirm functionality without rolling my own?

ntustison:
I'm guessing anything with ants*.sh constitutes a subset of "maintained" scripts.

Some of these scripts are exercised with various github examples (including data) that are linked to on the main ANTs github page.

gdevenyi:
@cookpa In the case of calls such as grep where there might not be a match, there is a workaround described at http://redsymbol.net/articles/unofficial-bash-strict-mode/#issues-and-solutions

Essentially, ``|| true`` handles those cases.

It's a bit of a flip in coding style, where you explicitly catch known "might fails" instead of explicitly checking return codes.

hjmjohnson:
Gabriel,

 

Thank you for taking this on.  I have been using this type of scripting approach for many years, and it certainly does make more robust and easier to use scripts.

 

Hans

gdevenyi:
Closed by #1771

PR #1974, opened 2026-05-20
BUG: Fix all shellcheck error-level findings in Scripts/ (refs #397)

hjmjohnson:
Eliminates all 83 `shellcheck -S error` findings (4 distinct rules) in `Scripts/`, fixing several latent bugs along the way. First step of the cleanup discussed in #397; warnings/info-level findings are not addressed here.

Each shellcheck rule is its own commit so individual fixes can be reverted or cherry-picked independently.

<details>
<summary>Per-commit breakdown</summary>

| Commit | Rule | Fixes | Notes |
|---|---|---|---|
| `BUG: SC1037 …` | SC1037 ×2 | `sliceBySliceOperation.sh` reading `$10` / `$11` | Bash parses bare `$10` as `${1}0`; the script silently stuffed garbage into PARAMETERS[5,6] whenever it was invoked with ≥10 args. Real bug. |
| `BUG: SC1087 …` | SC1087 ×6 | `antsMultivariateTemplateConstruction{,2}.sh`, `buildtemplateparallel.sh`, `directlabels.sh` | Two patterns: (a) `echo $files[1]` where `files` is a scalar from backtick `ls` (the `[1]` was dead text appended to the whole string) — index removed; (b) `echo "$TEMPLATES[$i] was not created"` where TEMPLATES is a real array — braces added so the right element is printed in the error message. |
| `BUG: SC2145 …` | SC2145 ×11 | `antsMultivariateTemplateConstruction{,2}.sh` | (a) `echo "prefix ${arr[@]}"` → `${arr[*]}` (echo wants one space-joined string). (b) `mv ${tmpdir}/selection/${TEMPLATES[@]} ${currentdir}/` was a real bug — the prefix only glued to the first element, so subsequent files were picked up from `cwd` instead of `${tmpdir}/selection/`. Replaced with an explicit `for` loop. |
| `BUG: SC2068 …` | SC2068 ×64 | 14 scripts | Mechanical `${arr[@]}` → `"${arr[@]}"` everywhere it was unquoted. Done with a perl in-place regex using negative lookbehind/ahead so already-quoted expansions are left alone. Protects against filenames-with-spaces being re-split by `IFS`. |

</details>

<details>
<summary>Verification</summary>

- `shellcheck -S error Scripts/*.sh` returns 0 lines after the four commits (was 83 lines before).
- `bash -n` syntax check passes on every modified file.
- No behavior change is intended in the SC2068 commit (pure quoting). The SC1087/SC2145 commits include three genuine bug fixes (positional-arg parse, error-message array index, `mv` cross-directory move) which alter observable behavior only in the broken-case direction.

</details>

<details>
<summary>What's NOT in this PR</summary>

- 497 shellcheck warnings (SC2206/SC2034/SC1083/SC2207/SC2124/SC2164/…) — these are the precondition cleanup needed before `set -euo pipefail` adoption can be safely attempted.
- `set -e` adoption itself — only 1 of 46 scripts uses strict mode today. That's a separate, per-script PR with maintainer regression testing per the discussion in #397.
- A CI lint gate. A reasonable next step is to add `shellcheck -S error Scripts/*.sh` to CI to keep the 0-error invariant.

</details>

Refs: #397


hjmjohnson:
@gdevenyi This was inspired by your issue #397.  The shell check tool lists several other warnings and suggestions, but this PR only addresses the perceived error cases (errors under conditions that may not exist, like spaces in filenames).

gdevenyi:
This will help a little bit but its a proper strict mode which I think is important for not having scripts run off and continue when errors are encountered.

hjmjohnson:
> This will help a little bit but its a proper strict mode which I think is important for not having scripts run off and continue when errors are encountered.

This is just one step to keep the PR not too big.

gdevenyi:
> This is just one step to keep the PR not too big.

Fair enough. I don't have review privileges here, but I approve :+1:

cookpa:
Thanks @hjmjohnson @gdevenyi I will review ASAP

Scripts/antsCorticalThickness.sh
cookpa, line 1: I ran Nick's 2D example with this script and output looks good

Scripts/antsMultivariateTemplateConstruction.sh
cookpa, line 1: Ran Nick's 2D syn template example on this, looks good

Scripts/antsAtroposN4.sh
cookpa, line 1: Ran the MalfLabelingExample using this

Scripts/antsLongitudinalCorticalThickness.sh
```diff
@@ -54,11 +54,11 @@ Usage:
               -p brainSegmentationPriors
               <OPTARGS>
               -o outputPrefix
-              \${anatomicalImages[@]}
+              \"${anatomicalImages[@]}"
 
 Example:
 
-  bash $0 -d 3 -e brainWithSkullTemplate.nii.gz -m brainPrior.nii.gz -p segmentationPriors%d.nii.gz -o output \${anatomicalImages[@]}
+  bash $0 -d 3 -e brainWithSkullTemplate.nii.gz -m brainPrior.nii.gz -p segmentationPriors%d.nii.gz -o output \"${anatomicalImages[@]}"
```
cookpa:
```suggestion
  bash $0 -d 3 -e brainWithSkullTemplate.nii.gz -m brainPrior.nii.gz -p segmentationPriors%d.nii.gz -o output "\${anatomicalImages[@]}"
```

Scripts/antsLongitudinalCorticalThickness.sh
```diff
@@ -54,11 +54,11 @@ Usage:
               -p brainSegmentationPriors
               <OPTARGS>
               -o outputPrefix
-              \${anatomicalImages[@]}
+              \"${anatomicalImages[@]}"
```
cookpa:
```suggestion
              "\${anatomicalImages[@]}"
```

cookpa (changes requested):
Needs a minor change in the longitudinal thickness script, to make usage format correctly.

cookpa (approved):

PR #1980, opened 2026-06-08
ENH: Diagnostic log metadata as separate CSV-stable line (supersedes #1978)

hjmjohnson:
Supersedes #1978 — addresses @hjmjohnson's CSV-preservation concern and the redundant-flush nits raised by Copilot, on top of @gdevenyi's original commit, rebased onto latest `master`.

<details>
<summary>What changed vs #1978</summary>

Three commits:

1. **`ENH: Add shrink factors, ...`** — @gdevenyi's original commit, unchanged, with `Co-Authored-By:` attribution preserved via cherry-pick.
2. **`ENH: Split per-level metadata into separate METADATA line`** — moves the `Key=Value` pairs (`ShrinkFactor=`, `SmoothingSigma=`, `MaxIterations=`, `FullScaleCCInterval=`) out of the `DIAGNOSTIC` header onto a separate `METADATA` line emitted once per level. The `DIAGNOSTIC` header reverts to pure CSV column names so downstream plotting tools see a stable schema. Prefixes mirror the existing pattern: `METADATA` / `XMETADATA` / `XXMETADATA` paired with `DIAGNOSTIC` / `XDIAGNOSTIC` / `XXDIAGNOSTIC`.
3. **`STYLE: Remove redundant std::flush before std::endl`** — `std::endl` already flushes; drop the redundant `std::flush`.

Example output (per level):

```
METADATA,Level=0,ShrinkFactor=4,SmoothingSigma=2,MaxIterations=100
DIAGNOSTIC,Iteration,metricValue,convergenceValue,ITERATION_TIME_INDEX,SINCE_LAST
2DIAGNOSTIC,     1, ...
2DIAGNOSTIC,     2, ...
```

</details>

<details>
<summary>Review threads from #1978 addressed</summary>

| Thread | Resolution |
|---|---|
| @hjmjohnson: "retain CSV output" | Header is now pure column names; values moved to `METADATA` line |
| Copilot: variable-schema header (3×) | Same — `METADATA` line carries the config |
| Copilot: redundant `std::flush << std::endl` (3×) | Removed across all three observer headers |

</details>

PR #1981, opened 2026-06-08
ENH: Print smoothing sigmas with fixed precision

cookpa:
Minor tweak to #1980, sometimes the default would print as an integer and sometimes as a float. 

Now print with fixed precision in scientific notation, so it will stay consistent between levels and maybe more readable for smaller voxels.


gdevenyi:
Great idea. I make some pretty gnarly smoothing sigmas when I programmatically generate them, and I would be fine if they were trimmed.

PR #2001, opened 2026-07-08
BUG: Fix principal-axis (Wahba) rotation with the proper polar factor

hjmjohnson:
Fixes the principal-axis (Wahba) rotation used by `antsAffineInitializer`, `antsAI`, and `ImageMath`. The proper-rotation step was wrong: two sites used an ad-hoc "flip the negative diagonal entries" correction and two used the raw `V·Uᵀ` with no correction at all (silently a reflection when `det = -1`). This replaces all four with the true special-orthogonal polar factor via a small gated helper, and adds the regression test the code path never had.

<details>
<summary>Root cause &amp; evidence</summary>

The attitude matrix `B = Σₖ outer(moving_axisₖ, fixed_axisₖ)` is **rank-deficient by construction** (a sum of outer products, rank `< n`). At the zero singular value the left/right null vectors have independent sign, so `V·Uᵀ` is not a well-defined rotation, and the ad-hoc diagonal-sign fix does not repair it.

Measured over 2000 consistent Wahba problems with a known ground-truth rotation:

| Correction | recovers the rotation |
|---|---|
| old heuristic (both `vnl_svd` and Eigen) | **~52%** (fails ~48%) |
| proper polar factor `R = U·diag(1,…,det(U·Vᵀ))·Vᵀ` | **100%**, backend-invariant |

</details>

<details>
<summary>What changed</summary>

- New `Utilities/antsNearestRotation.h` — `ants::NearestRotation(A)` returns the SO(n) polar factor minimizing `‖R − A‖`. It prefers the Eigen-backed `itk::Math::SVD` when the ITK in use provides it and falls back to `vnl_svd` otherwise; the result is identical either way (the correction cancels the singular-vector sign freedom).
- `antsAffineInitializer`, `antsAI` (both Wahba sites), `ImageMath` now call `ants::NearestRotation` instead of the raw product / diagonal-sign heuristic.

</details>

<details>
<summary>Test plan</summary>

Adds `antsWahbaRotationTest` (fixture-free, deterministic): for several known 2-D and 3-D rotations it builds `B` exactly as the initializers do and asserts recovery of `R`, orthogonality (`RᵀR = I`), and `det(R) = +1`. Verified locally: all cases pass; the previous raw-`V·Uᵀ` approach fails every case, so the test genuinely guards the fix. `git clang-format` against `main` reports no changes to the modified lines.

</details>

### DIFF
--- a/Examples/ImageMath_Templates.hxx
+++ b/Examples/ImageMath_Templates.hxx
@@ -13,6 +13,7 @@
 =========================================================================*/
 
 #include "antsUtilities.h"
+#include "antsNearestRotation.h"
 #include <algorithm>
 #include <vnl/vnl_inverse.h>
 #include "itkTransformFileReader.h"
@@ -13065,8 +13066,8 @@ PatchCorrelation(itk::NeighborhoodIterator<TImageType> GHood,
     //        outer_product( evec2_primary , evec1_primary ) * wt0 / evsum;
     B = outer_product(evec2_2ndary, evec1_2ndary) + outer_product(evec2_primary, evec1_primary);
   }
-  vnl_svd<RealType>    wahba(B);
-  vnl_matrix<RealType> A_solution = wahba.V() * wahba.U().transpose();
+  // Inverse of the nearest rotation to B (proper, backend-invariant Wahba solution).
+  vnl_matrix<RealType> A_solution = ::ants::NearestRotation(B).transpose();
   //   " << vnl_determinant<RealType>(  wahba.U()) << std::endl;
   // now rotate the points to the same frame and sample the neighborhoods again
   for (unsigned int ii = 0; ii < Gsz; ii++)
--- a/Examples/TestSuite/CMakeLists.txt
+++ b/Examples/TestSuite/CMakeLists.txt
@@ -57,6 +57,14 @@ MakeTestDriverFromANTSbinary(antsApplyTransformsTest antsApplyTransformsTest.cxx
 MakeTestDriverFromANTSbinary(compareTwoTransformsTest compareTwoTransformsTest.cxx compareTwoTransforms)
 MakeTestDriverFromANTSbinary(simpleSynRegistrationTest simpleSynRegistrationTest.cxx simpleSynRegistration)
 
+## Wahba principal-axis rotation regression test: guards the proper
+## ants::NearestRotation used by antsAffineInitializer / antsAI. Fixture-free and
+## deterministic; would fail on the previous ad-hoc det-correction heuristic.
+add_executable(antsWahbaRotationTest antsWahbaRotationTest.cxx)
+target_include_directories(antsWahbaRotationTest PRIVATE ${CMAKE_SOURCE_DIR}/Utilities)
+target_link_libraries(antsWahbaRotationTest ${ITK_LIBRARIES})
+add_test(NAME antsWahbaRotationTest COMMAND antsWahbaRotationTest)
+
 
 ###
 #  Perform testing
--- a/Examples/TestSuite/antsWahbaRotationTest.cxx
+++ b/Examples/TestSuite/antsWahbaRotationTest.cxx
@@ -1,0 +1,125 @@
+/*=========================================================================
+ *
+ *  Regression test for the Wahba principal-axis rotation used by
+ *  antsAffineInitializer / antsAI. Those tools build an attitude matrix
+ *  B = sum_k outer(moving_axis_k, fixed_axis_k) and recover the aligning
+ *  rotation from its SVD. B is rank-deficient by construction, so the proper
+ *  special-orthogonal polar factor (ants::NearestRotation) must be used; an
+ *  ad-hoc "flip negative diagonal" correction fails ~48% of the time. This
+ *  test constructs consistent problems with a known ground-truth rotation and
+ *  asserts exact recovery, orthogonality and det = +1. It is fixture-free and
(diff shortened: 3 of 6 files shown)


hjmjohnson:
## Correctness & backend-equivalence proof (harness details)

**Verdict:** the fix is **correctness-motivated, not performance-motivated** — `NearestRotation` is the same `O(SVD)` cost as the code it replaces. The harness proves two things: (1) the proper polar factor **recovers the true rotation in 100% of cases** where the old heuristic failed ~48% of the time, and (2) the result is **identical across SVD backends** (`vnl_svd` ≡ Eigen `itk::Math::SVD`), so the optional `itk::Math::SVD` path changes nothing numerically.

**Environment**

| | |
|---|---|
| OS / CPU | macOS (Darwin 25.5.0), Apple M3 Max, 16 cores |
| Compiler | Apple clang 21.0.0 |
| `itk::Math::SVD` provider | ITK `5e836abec33` (vnl_svd = ITK's vendored VXL) |
| Harness | single-threaded, deterministic (fixed rotations / fixed RNG seed) |

### 1. Old heuristic vs proper polar factor — 2000 consistent Wahba problems (`wahba4.cxx`)

Each trial: a known ground-truth rotation `R`, fixed axes `eₖ`, moving axes `R·eₖ`, then `B = Σ outer(R·eₖ, eₖ)` (rank-deficient, exactly as `antsAffineInitializer` builds it). "misaligned" = the recovered transform fails to map the axes back (`‖A·eₖ − R·eₖ‖ > 1e-6`).

| correction | mean align error | **misaligned / 2000** |
|---|---|---|
| ANTs heuristic — `vnl_svd` | 1.2480 | **964 (48%)** |
| ANTs heuristic — Eigen | 1.2875 | **989 (49%)** |
| **proper polar factor — `vnl_svd`** | **0.0000** | **0** |
| **proper polar factor — Eigen** | **0.0000** | **0** |

The heuristic is broken regardless of backend (`itk` worse than `vnl` in 403/2000, better in a comparable share — pure noise from which arbitrary null-space sign it latches onto). The proper factor recovers every case, both backends.

### 2. Where the backends *disagree*, Eigen was the correct one (`wahba3.cxx`, first 5 reps)

```
rep   |Av-Ai|     align_vnl   align_itk    |Av-Rt|    |Ai-Rt|
0     4.9e-16     7.4e-16     7.4e-16      5.5e-16    8.6e-16   both correct
1     4.5e-16     1.56        1.56         1.59       1.59      both wrong (heuristic)
2     1.60        2.67        2.3e-16      2.77       6.4e-16   only itk correct
3     1.30        2.14        1.9e-15      2.25       2.0e-15   only itk correct
4     4.6e-16     2.26        2.26         2.62       2.62      both wrong (heuristic)
```

This is why "the two backends differ" was never evidence of an Eigen bug — the divergence is the *heuristic* amplifying an arbitrary sign; the proper factor removes it entirely.

### 3. `NearestRotation` recovery + backend-invariance (`nrot.cxx`)

```
NearestRotation recovers Rt & backend-invariant: 1000/1000 PASS
```
(`‖R − R_true‖`, `‖RᵀR − I‖`, `|det(R) − 1|` all `< 1e-9`; `vnl` vs `itk` agree to machine precision.)

### 4. The committed regression test catches the regression

`antsWahbaRotationTest` asserts recovery + orthogonality + `det = +1` on known 2-D/3-D rotations. The pre-fix raw `V·Uᵀ` fails **every** case (`‖R − R_true‖` = 1.2–2.8), so the test is a genuine guard, not a tautology.

<details>
<summary>Reproducer</summary>

Harnesses live in `~/src/ITK/.devlocal/svd-category-a-equivalence/` (built against an `itk::Math::SVD`-enabled ITK via `find_package(ITK)`):

```
ninja -C build wahba3 wahba4 nrot
./build/wahba4   # 2000-problem statistics (table 1)
./build/wahba3   # per-rep backend comparison (table 2)
./build/nrot     # recovery + backend-invariance (table 3)
```
`wahba4.cxx` replicates ANTs' exact heuristic (`inverse(V·Uᵀ)` + flip-negative-diagonal) alongside the proper factor; `nrot.cxx` exercises the shipped `ants::NearestRotationImpl` formula.
</details>

ntustison:
Really interesting.  Thanks @hjmjohnson

hjmjohnson:
@ntustison @stnava I came across this while trying to update the SVD infrastructure in ITK from vnl -> Eigen.  FULL DISCLOSURE: The analysis and results are AI-driven (as you probably guessed).  

My hypothesis:  vnl_svd should be very similar to Eigen SVD results, but this case in my testbed failed that hypothesis, and upon further investigation, this branch was created.

This could use a good review of the math and intended behaviors to ensure that this PR is the appropriate path to take.

gdevenyi:
This is great, I was trying to use antsAI for something and found it consistently not working well, I'll have to revisit it with these proposed changes.

ntustison (approved):
Thanks again, @hjmjohnson .  I doubt I can offer anything beyond the reasoning provided  but it looks good to me.

cookpa:
Thanks for this, @hjmjohnson. If I understand correctly, this affects uses with the `--align-principal-axes` option?

gdevenyi:
> Thanks for this, @hjmjohnson. If I understand correctly, this affects uses with the `--align-principal-axes` option?

@hjmjohnson can you confirm?

cookpa:
Looks like `--align-blobs` also triggers a call to the corrected rotation.

PR #2011, opened 2026-07-29
COMP: Use minimal CPU flags by default

### LINKED ISSUE #2007: Update default compiler flags
We are still using hard-coded "corei7", which may have platform compatibility issues

https://github.com/InsightSoftwareConsortium/ITK/issues/2634

cookpa:
Should make life easier for distribution. Users can add optimizations for local builds if they like

Fixes #2007


cookpa:
Removed other outdated stuff, now pretty close to current ITK version

gdevenyi:
Maybe also want to set -march to a sensible default
image

cookpa:
A baseline might be a good idea. Looks like this is being considered at ITK also https://github.com/InsightSoftwareConsortium/ITK/issues/6749

hjmjohnson (approved):

hjmjohnson:
@cookpa This is a good step in the right direction.  I think any changes from this should be a separate commit and PR.

cookpa:
Going to merge for now, just using platform defaults. Happy to revisit in future PRs.

PR #2018, opened 2026-08-12
BUG: Rethink "strict mode" in template scripts

### LINKED ISSUE #2017: IFS - antsMultivariateTemplateConstruction2.sh
### Operating system and version

Almalinux 8

### CPU architecture

x86_64 (PC, Intel Mac, other Intel/AMD)

### ANTs code version

2.6.5

### ANTs installation type

Downloaded Github release binaries

### Summary of the problem

The commit https://github.com/ANTsX/ANTs/commit/5debf1487b625d95db1f949bff7728857b8dacad introduced the line IFS=$'\n\t' into antsMultivariateTemplateConstruction2.sh. I'm guess this was intended to deal with spaces in file names, but it is only a guess. Unfortunately it also breaks the calls to the waitFor perl scripts:
```
waitForSlurmJobs.pl 1 60 $jobIDs
```
The $jobIDs should be a space separated string of numeric job ids: "1 2 3 4 5 6", bash converts these to individual arguments, the perl script adds these to a list. With IFS set the $jobIDs is passed as a single string " 1 2 3 4 5 6", perl adds this to a single item list. The resulting call to squeue will always fail:
```
 --------------------------------------------------------------------------------------
 Starting ANTS rigid registration on SLURM cluster. Submitted 20 jobs 
--------------------------------------------------------------------------------------
  Waiting for 1 jobs:  6715400 6715401 6715402 6715403 6715404 6715405 6715406 6715407 6715408 6715409 6715410 6715411 6715412 6715413 6715414 6715415 6715416 6715417 6715418 6715419
squeue: error: Unrecognized option: 6715401
```

### Commands to reproduce the problem.

The easiest way to demonstrate is via two cut down scripts:

```
cat waitFor.pl 
#!/usr/bin/perl -w

use strict;

my ( $verbose, $delay, @jobIDs ) = @ARGV;

print "scalar jobIDs: " . scalar @jobIDs . "\n";
print "jobIDs[0]: -" . $jobIDs[0] . "-\n";
exit;
% 
```

```
% cat bash_script.pl 
#!/bin/bash
IFS=$'\n\t'

jobIDs=" 1 2 3 4 5 6"
./waitForSlurmJobs.pl 1 60 $jobIDs
% 
```

```
% ./bash_script.pl 
scalar jobIDs: 1
jobIDs[0]: - 1 2 3 4 5 6-
% 
```

Comment out the IFS line in the bash script and you get the correct behaviour:
```
% ./bash_script.pl
scalar jobIDs: 6
jobIDs[0]: -1-
%
```

### Output of the command with verbose output.

N/A

### Data to reproduce the problem

N/A

cookpa:
Undo the global custom IFS

Use local IFS where needed

More extensive use of arrays and quoting.

From Codex:

Updated antsMultivariateTemplateConstruction2.sh
Key changes:
* Removed the global custom IFS.
* Preserved input arguments with IMAGESETARRAY=("$@").
* Used locally scoped IFS only for CSV, modality, and level parsing.
* Replaced ls-based filename collection with globbed arrays.
* Quoted direct filesystem and command arguments.
* Shell-escaped paths embedded in generated job scripts using printf %q.
* Converted scheduler job IDs to arrays.
* Guarded all eight waitFor*.pl calls with if ! ..., ensuring warnings print under set -e.

Updated antsMultivariateTemplateConstruction.sh
Changes include:
* Added set -euo pipefail and shopt -s extglob.
* Fixed variables that strict mode exposed as potentially unset.
* Initialized the default template name and linear transform parameters.
* Made help handling safe when no arguments are provided.
* Replaced unsafe filename strings and ls parsing with arrays.
* Preserved command-line and CSV filenames containing spaces.
* Quoted direct filesystem and program arguments.
* Shell-escaped filenames written into generated job scripts.
* Converted scheduler job IDs to arrays.
* Guarded all eight waitFor*.pl calls so scheduler warnings appear under set -e.
* Replaced arithmetic commands such as ((count++)) and let, which can unexpectedly return failure under set -e.
* Corrected -x to assign XGRIDOPTS=$OPTARG.
* Added a CPU-count fallback for strict-mode compatibility.

# Description

### DIFF
--- a/Scripts/ANTSpexec.sh
+++ b/Scripts/ANTSpexec.sh
@@ -153,7 +153,7 @@ while getopts j:rh OPT; do # "j:" waits for an argument "h" doesnt
 done
 
 # Main program
-echo Using max $MAX_NPROC parallel threads
+echo Using max $MAX_NPROC parallel processes
 
 if [ $MAX_NPROC -eq 1 ] ; then
 echo " Dont use pexec to run 1 process at a time. "
--- a/Scripts/antsMultivariateTemplateConstruction.sh
+++ b/Scripts/antsMultivariateTemplateConstruction.sh
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+set -euo pipefail
+
+# Use extended globbing to allow for more flexible pattern matching
+shopt -s extglob
+
 VERSION="0.0.0"
 
 # trap keyboard interrupt (control-c)
@@ -17,9 +22,9 @@ XGRID=waitForXGridJobs.pl
 SLURM=waitForSlurmJobs.pl
 
 fle_error=0
-for FLE in $ANTS $WARP $N4 $PEXEC $SGE $XGRID $PBS $SLURM
+for FLE in "$ANTS" "$WARP" "$N4" "$PEXEC" "$SGE" "$XGRID" "$PBS" "$SLURM"
   do
-    if ! command -v $FLE &> /dev/null
+    if ! command -v "$FLE" &> /dev/null
       then
         echo
         echo "--------------------------------------------------------------------------------------"
@@ -55,7 +60,7 @@ Compulsory arguments (minimal command line requires SGE cluster, otherwise use -
           line is the location of the input image.  One can also specify more than
           one file for each image for multi-modal template construction (e.g. t1 and t2).
           For the multi-modal case, the templates will be consecutively numbered (e.g.
-          ${OUTPUTPREFIX}template0.nii.gz, ${OUTPUTPREFIX}template1.nii.gz, ...).
+          \${OUTPUTPREFIX}template0.nii.gz, \${OUTPUTPREFIX}template1.nii.gz, ...).
 
 NB: All images to be added to the template should be in the same directory, and this script
 should be invoked from that directory.
@@ -84,7 +89,7 @@ Optional arguments:
      -i:  Iteration limit (default 4) -- iterations of the template construction (Iteration limit)*NumImages registrations.
 
      -j:  Number of cpu cores to use.  For pexec option (default 2; requires "-c 2")
-          or Slurm's --cpu-per-tasks.  
+          or Slurm's --cpu-per-tasks.
 
      -k:  Number of modalities used to construct the template (default 1)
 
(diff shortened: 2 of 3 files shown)


gdevenyi:
Great work. I've been meaning to look into this for a while now but its so invasive and this is a big pipeline.

cookpa:
I updated the example to be more user friendly with current ANTs

https://github.com/ntustison/TemplateBuildingExample/pull/10

cookpa:
So I went down a bit of a rabbit hole here, because the scripts did not work for me with slurm. 

My latest commit undoes #1956, the recent repurposing of -j to set slurm cores. Instead, I've added `-T` to explicitly set the number of threads, with sensible fallbacks from the environment ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS or default values, like we do in the pairwise registration scripts. This should put a stop to people blowing up their machines with massive numbers of threads. It breaks backwards compatibility, but only since March of this year.

I believe the issues #1955 date way back to when ANTSPATH was removed. A stray "--export" was left in the sbatch command, which was later merged with the subsequent '$QSUBOPTS' to produce `--export=${QSUBOPTS}`. This sends "$QSUBOPTS" to the job as a list of environment variables, not cluster commands, which would explain why the user in #1955 was unable to successfully modify the sbatch call. Also, using `--export` replaces the usual default (inherit the submission environment) and leads to PATH and multi-threading variables not being read.

Separately, the script was inconsistent about where $QSUBOPTS appeared in the command line. And there were several hard-coded options like `--cpus-per-task=1` that override options added at run time with `-p`. Thus even if slurm users set `ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS` correctly, they would never get more than one CPU to run the job (the CPUs issue was fixed by #1956, but only for slurm, not SGE / PBS).

For pexec execution, `-j` remains the way to control the number of parallel processes, ie the number of registrations that run in parallel. It is, as it previously was, distinct from the number of threads, controlled by either `-T` or `ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS`.

New defaults:

* -c 0 -> min(8,cpu_count) threads, serial execution
* -c 2 -> 1 thread per process, parallel processes controlled by -j
* -c <1=SGE/4=PBS/5=slurm> -> 1 thread per process, parallel processes controlled by job scheduler

If `ITK_GLOBAL_DEFAULT_NUMBER_OF_THREADS` is set to a positive value, it is used, so for pexec this preserves previous behavior. One can also use `-T` on the command line, which takes precedence over the environment.

The number of threads is prepended to the job scripts, so it won't get lost if users add `--export` in the prepend string passed with `-p`.

tl;dr, if you're using slurm, the script prepend option should now be functional and you can control threads-per-process with `-T`, and this will be correctly propagated to the job, and `--cpus-per-task` will also be set to the same value.

cookpa:
For SGE / PBS, I removed some site-specific things and hard-coded single thread options that would likewise override options passed with `-p`. In the usage documentation, I note that SGE and PBS users need to use `-p` to pass parallel environment options, in order to reserve multiple CPUs when multi-threading.

cookpa:
There are other issues but I will save them for another PR. @stnava @ntustison any objections to merging this? It should not break anything unless you are using the `-j` options for slurm.

ntustison:
Yeah, fine with me.  THanks @cookpa

PR #2019, opened 2026-08-19
COMP: Prefer the tiered itk::bridge::Math spelling for ITK's Eigen numerics

hjmjohnson:
Keeps ANTs on ITK's Eigen-backed numerics now that InsightSoftwareConsortium/ITK#6768
has **merged**, moving the convenience numerics to `itk::bridge::Math`
(InsightSoftwareConsortium/ITK#6620). Chained `__has_include` guards keep both the
post-move and pre-move ITK on the Eigen path, so this builds against either vintage.

Affects `Utilities/antsNearestRotation.h` (nearest-rotation / Wahba solve),
`Tensor/TensorFunctions.h`, `Utilities/antsSCCANObject.h`, and
`Utilities/itkVectorFieldGradientImageFunction.hxx`.

<details>
<summary>Why this is needed even though nothing breaks</summary>

The pre-existing guards were bare `__has_include` probes of the pre-move header:

```cpp
#if __has_include(<itkMathLDLT.h>)
#  include <itkMathLDLT.h>
#endif
```

After the ITK rename that probe simply fails. There is **no compile error** — the
capability macro stays undefined and the code silently falls back to
`vnl_matrix_inverse` / `vnl_svd`, with nothing in any log to say so.

Chaining the probe keeps either ITK vintage on the Eigen path:

```cpp
#if __has_include(<itkBridgeMathLDLT.h>)
#  include <itkBridgeMathLDLT.h>
#elif __has_include(<itkMathLDLT.h>)
#  include <itkMathLDLT.h>
#endif
```

Because both arms are retained, this is safe against pre-move and post-move ITK alike.

</details>

<details>
<summary>Test plan — built and measured against merged ITK</summary>

Validated against ITK `main` at `313b5234d52` (i.e. after #6768 merged), installed
from a dedicated build with the remote modules ANTs requires
(`GenericLabelInterpolator`, `AdaptiveDenoising`, `ITKReview`, `ITKMINC`,
`ITKIOTransformMINC`, `ITKTBB`).

**1. Guard-arm probe.** A syntax-only compile of each guard against the installed
ITK, with one `#error` per arm, reports which branch the preprocessor actually takes:

| Guard | Files | Result |
|---|---|---|
| `itkBridgeMathLDLT.h` chain | `TensorFunctions.h`, `antsSCCANObject.h`, `itkVectorFieldGradientImageFunction.hxx` | `ARM=BRIDGE(eigen)` |
| `itkBridgeMathSVD.h` chain | `antsNearestRotation.h` | `ARM=BRIDGE(eigen)` |

Neither falls through to the `vnl` arm. This matters because a successful build is
consistent with *all three* arms — the `vnl` fallback compiles fine, it just quietly
abandons Eigen. The probe is what distinguishes them.

**2. Full build.** `ninja` over the whole project: **417/417 targets, 0 errors,
0 `FAILED` lines.**

Not run: the ANTs test suite. This change alters which numerical backend is selected,
not the algorithms, and the Eigen/vnl equivalence was measured previously
(<=2.6e-15 on random SPD 3x3 tensors).

</details>

### DIFF
--- a/Tensor/TensorFunctions.h
+++ b/Tensor/TensorFunctions.h
@@ -20,20 +20,24 @@
 // diffusion-tensor inverses below, when a new-enough ITK provides it. Gated on a
 // compile-time capability check so this still builds against the currently pinned
 // ITK (which takes the legacy vnl_matrix_inverse path).
-#if __has_include(<itkMathLDLT.h>)
+#if __has_include(<itkBridgeMathLDLT.h>)
+#  include <itkBridgeMathLDLT.h>
+#elif __has_include(<itkMathLDLT.h>)
 #  include <itkMathLDLT.h>
 #endif
 
 namespace matHelper
 {
-// Inverse of a SYMMETRIC matrix. With itk::Math::SolveSymmetric available, build
+// Inverse of a SYMMETRIC matrix. With itk::bridge::Math::SolveSymmetric available, build
 // the inverse column-by-column via LDLT solves (DT X = I); otherwise fall back to
 // vnl_matrix_inverse. Validated equivalent to <=2.6e-15 on random SPD 3x3 tensors
 // (see .devlocal/tensor-solve-prototype).
 inline vnl_matrix<double>
 SymmetricInverse(const vnl_matrix<double> & A)
 {
-#ifdef ITK_MATH_HAS_SOLVE_SYMMETRIC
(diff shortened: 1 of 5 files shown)


ntustison (approved):
Thanks @hjmjohnson

cookpa (commented):
Looks good

cookpa:
Thanks for this @hjmjohnson

PR #2021, opened 2026-08-22
BUG: Remove accidentally committed CTest artifacts

hjmjohnson:
Removes two transient CTest outputs that were accidentally committed to the source tree — `Examples/Testing/Temporary/CTestCostData.txt` and `LastTest.log` — and adds a `.gitignore` rule so they cannot come back.

<details>
<summary>Why these are accidental, not intentional fixtures</summary>

Both files entered the tree in d9b0d680 ("ENH: Quiet option."), a commit whose subject and remaining ten files are unrelated source changes — `antsRegistration.cxx`, `antsApplyTransforms.cxx`, `antsMotionCorr.cxx`, `itkantsRegistrationHelper.h/.hxx`, and others. The signature of a `git add -A` sweep.

Supporting evidence:

- **Contents are empty artifacts.** `CTestCostData.txt` contains only `---`. `LastTest.log` records a test run with no tests: `Start testing: Nov 12 21:41 PST` / `End testing: Nov 12 21:41 PST`.
- **Never touched since.** One commit in their entire history — the one that added them in 2015.
- **Unreferenced.** No `*.cmake`, `CMakeLists.txt`, `*.sh`, or `*.py` in the repo mentions either path.
- **Regenerated anyway.** CTest rewrites both into the build tree's `Testing/Temporary` on every invocation, so the tracked copies are shadowed and never read.

</details>

<details>
<summary>Why the ignore pattern needs the leading <code>**/</code></summary>

The rule added is:

```
**/Testing/Temporary/
```

A gitignore pattern containing a slash anywhere other than at the end is **anchored to the repository root**. Writing `Testing/Temporary/` would therefore match only `<root>/Testing/Temporary/` and would *not* match `Examples/Testing/Temporary/` — the very path this PR removes. Verified with `git check-ignore -v` against a probe file at that depth before and after the fix.

Note also that `.gitignore` has no effect on files already tracked, which is why these two survived in the tree; untracking them is what makes the rule effective.

</details>


cookpa (approved):
Thanks @hjmjohnson

PR #2026, opened 2026-09-05
BUG: KellyKapowski smoothing usage said mm but was voxel units

cookpa:
The option `-l` was said to be "the Gaussian smoothing of the hit and total images.  Default = 1.0 mm.", but it was actually used in two places.

It was used as gradient smoothing sigma, in mm, and then as the total / hit smoothing *variance* in voxels^2.

Fix by adding a parameter `-j` to set the gradient smoothing sigma, in mm, and setting `-l` to be in mm^2, not voxel units.

BUG: Convergence reporting overflowed due to float typing

This was identified by Codex. It doesn't affect typical usage (fixed iterations).

- Used double energy accumulators and itk::SizeValueType counters to prevent loss of integer precision during convergence reporting.
- Added zero-count guards before normalization.
- Updated help text and diagnostic output.

### DIFF
--- a/Examples/KellyKapowski.cxx
+++ b/Examples/KellyKapowski.cxx
@@ -269,6 +269,17 @@ DiReCT(itk::ants::CommandLineParser * parser)
   }
 
   //
+  // gradient smoothing sigma
+  //
+  typename itk::ants::CommandLineParser::OptionType::Pointer gradientSmoothingSigmaOption =
+    parser->GetOption("gradient-smoothing-sigma");
+  if (gradientSmoothingSigmaOption && gradientSmoothingSigmaOption->GetNumberOfFunctions())
+  {
+    direct->SetGradientSmoothingSigma(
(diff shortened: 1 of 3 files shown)


ntustison:
Thanks @cookpa .  Yeah, this is something that I've known about for quite some time and I think it might go all the way back to the original KellySlater code.  However, since there's really no parameter variation in actual usage, I never fixed it.    But thanks for this.

cookpa:
I came across this while investigating the influence of spatial resolution on thickness estimates. Upsampled versions of the same segmentation report greater thickness. This may be somewhat expected but I think it is exacerbated by implicit changes in smoothing scale.

The velocity smoothing variance is also in voxels, I didn't change this because it is documented as being in voxels.

ntustison (approved):

cookpa:
Thinking about this some more, I think since the defaults are so heavily used, the correct thing is to correct the documentation and not the code. This maintains backwards compatibility with defaults.

